PR #10, opened 2025-09-01
(CM-413) Output organisation name in host editions

pezholio:
Before, we linked out to an organisation in Whitehall, which we can’t do now. The usefulness of this is probably limited anyway, so for now, let’s just show the organisation’s name.


edavey (approved):
👍

PR #20, opened 2025-09-02
(CM-416) Add `assets.prefix` to the Asset config

pezholio:
Apparently we need this for the app to come up cleanly


edavey (approved):
🤞

pezholio:
Ah ha, think I've got it

PR #23, opened 2025-09-03
Add ADR for use of CHANGELOG file

edavey:


edavey:
> I'm on the fence about this for a couple of reasons:
> 
> 1. We don't want to make this overly onerous
> 2. Equally, if we do want to do this - we don't want it to be something that can easily be ignored
> 3. As we're continuously deploying, it'll more than likely be one Changelog entry per release, which might be counterintuitive
> 
> Interested to hear your thoughts 👍

hmm... I'm in favour of:

- continuous deployment for minimising risk, and am in favour of
- the discipline of having a changelog entry with each bugfix or feature merge

I'm not sure how to combine the two, but I expect a bit of research will throw some light on it.

For now I will find another way of adding the Govspeak documentation I want to record.

edavey:
closing for now

PR #24, opened 2025-09-03
(CM-415) Update documentation

pezholio:
This updates the README, as well as moving the relevant pieces of documentation from the `/docs` directory.


edavey (approved):
how would you feel about putting the decision records in `docs/architecture/decisions` to be more explicit and score a small win in the battle against acronymns?

PR #39, opened 2025-09-09
(CM-424) Add import task

pezholio:
We have some JSON files (one for each model), which we intend to upload to the root application directory. Once this is done, we can run this task. The task runs in a transaction, so if there are any errors (such as a file not being found, or a validation error), any changes are rolled back.

To run this import, we'll copy the files to a content block manager pod like so:

```
kubectl cp FILENAME POD_NAME:/tmp/FILENAME
```

Once all the files are added, we can run the command:

```
kubectl exec -it POD_NAME -- rake content_block_manager:import
```


edavey (approved):
looks, good 👍

PR #40, opened 2025-09-09
(CM-415) Organisation tweaks

pezholio:
This adds a couple of tweaks for Organisations - firstly, we ensure they are listed alphabetically. Secondly, we add the `select-with-search` JS, so the Organisation list is easier to work with (we already have this in the old CBM, but the lack of Javascript prevented this from working properly)

I also added `mutex_m` to the Gemfile, so I can run the tests in Rubymine.

### Before

image

image

### After

image

image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -4,6 +4,7 @@
 //= require govuk_publishing_components/components/checkboxes
 //= require govuk_publishing_components/components/copy-to-clipboard
 //= require govuk_publishing_components/components/govspeak
+//= require govuk_publishing_components/components/select-with-search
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/application.js
```diff
@@ -4,6 +4,7 @@
 //= require govuk_publishing_components/components/checkboxes
 //= require govuk_publishing_components/components/copy-to-clipboard
 //= require govuk_publishing_components/components/govspeak
+//= require govuk_publishing_components/components/select-with-search
```
edavey, line 7: is it worth including some info or a link to explain what behaviour this JS library adds? (e.g. I have no idea!)

PR #44, opened 2025-09-10
(CM-424) Reset Primary Key sequence of all tables

pezholio:
There are some join tables that get written to that we need to reset the keys of too. To be on the safe side, let’s reset the sequence for all tables, not just the ones for the models we’re acting on


pezholio:
Yep, I'll bin all this code once we're done with it 👍

PR #50, opened 2025-09-15
(CM-126) ADR for reordering work

pezholio:
[Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/CM/boards/134?selectedIssue=CM-126)

This lays out the proposed approach we propose to take to allow items in blocks to be reordered.

### DIFF
--- a/docs/architecture/decisions/0004-allow-order-to-be-specified.md
+++ b/docs/architecture/decisions/0004-allow-order-to-be-specified.md
@@ -1,0 +1,128 @@
+# 4. Allow order of items to be specified in blocks
+
+Date: 2025-09-15
+
+## Status
+
+Accepted
+
+## Context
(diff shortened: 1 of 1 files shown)


edavey:
Should we include a note in "consequences" about the possible need or expectation that custom ordering would be reflected in the workflow review step (`/editions/{editionId}/workflow/review`). e.g. if I've defined a custom order on my "contact" block of

```json
"order": [
  "addresses.main-address", 
  "telephones.main-telephone",
  "email_addresses.main-email", 
  "email_addresses.secondary-email", 
  "telephones.secondary-telephone",
  "addresses.secondary-address"
], 
```

do we need to apply that custom order when we render the "review" step?:

review_contact_methods_in_custom_order

PR #55, opened 2025-09-17
Disable Dependabot

samsimpson1:
Disable Dependabot for all dependencies while the npmjs security incident is ongoing


.github/dependabot.yml
```diff
@@ -1,12 +1 @@
...
-  directory: "/"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
-- package-ecosystem: github-actions
-  directory: "/"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
+---
```
pezholio, line 1: At the moment, this is causing Github actions to complain. It might be better to set `open-pull-requests-limit` to `0` to disable it. I'm guessing we need to add an entry for `npm` too. Something like this should do the trick:

```
version: 2
updates:
- package-ecosystem: bundler
  directory: "/"
  schedule:
    interval: daily
  open-pull-requests-limit: 0
- package-ecosystem: github-actions
  directory: "/"
  schedule:
    interval: daily
  open-pull-requests-limit: 0
- package-ecosystem: npm
  directory: /
    schedule:
      interval: daily
  open-pull-requests-limit: 0
```

dj-maisy (approved):

PR #58, opened 2025-09-18
Don’t render a link when Host Edition has no base_path

pezholio:
We recently saw an error when embeddding a content block within another content block, as content block’s don’t have a base_path. We don’t expect this to happen in real life too much, but let’s add a guard in case someone does it.


edavey (approved):
👍

PR #60, opened 2025-09-18
Add custom error pages

pezholio:
This adds custom error pages by setting `config.exceptions_app` to `routes`, as well as setting up the routes.

We already had the error views in the app from the migration, but I’ve removed the setting of the `product_name`, which isn’t necessary, as well as adding a view from the `unprocessable_content` error, which wasn’t there

I've also moved the views out of the `admin` namespace, which we don't make use of.


edavey (approved):
nice!

PR #70, opened 2025-09-24
CM-374: Describe Govspeak-enabled fields in Contact

edavey:
This cucumber feature verifies that the expected fields within the "Contact" 
content block are indeed Govspeak-enabled.

We need to ensure that our stubbed schema includes the
properties which are (in `/config/content_block_manager.yml`)
configured as `govspeak_enabled: true` -- and then we can assert
that the expected textareas are indeed wrapped in the
"govspeak editor".

This should prevent inadvertent regressions, e.g. when the the config is edited.

We also tweak the spacing around the "Govspeak supported" guidance:

style-margins_govspeak_textarea

### DIFF
--- a/features/step_definitions/govspeak_steps.rb
+++ b/features/step_definitions/govspeak_steps.rb
@@ -6,6 +6,10 @@ Then("I see that the pension rate description is Govspeak-enabled") do
   expect_to_see_a_govspeak_enabled_textarea_for_id("edition_details_rates_description")
 end
 
+Then("I see that the contact address description is Govspeak-enabled") do
+  expect_to_see_a_govspeak_enabled_textarea_for_id("edition_details_addresses_description")
+end
+
 def expect_to_see_a_govspeak_enabled_textarea_for_id(id)
   assert(
     page.has_selector?(".app-c-govspeak-editor ##{id}"),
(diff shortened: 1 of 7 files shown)


features/step_definitions/govspeak_steps.rb
```diff
@@ -6,6 +6,10 @@
   expect_to_see_a_govspeak_enabled_textarea_for_id("edition_details_rates_description")
 end
 
+Then("I see that the contact address description is Govspeak-enabled") do
```
pezholio, line 9: I wonder if it might be a bit more DRY to have a regex here? Something like:

```
Then(/^I see that the ([^"]*) is Govspeak-enabled$/) do |field|
  id = case field
         when "contact address description"
            "edition_details_addresses_description"
         ... etc
    end
    expect_to_see_a_govspeak_enabled_textarea_for_id(id)
end
```

pezholio (commented):
LGTM - just one suggestion to DRY things up

features/step_definitions/govspeak_steps.rb
edavey: good idea, thks: it's nice to have all the IDs listed together

pezholio (approved):

PR #80, opened 2025-09-29
(CM-508) Add preview to review screen when adding/editing a contact

pezholio:
This allows users to preview their contacts on the review screen as well as the other subschema-related screens.

### Screenshots

image

image

image

image


edavey (approved):
👍

PR #82, opened 2025-09-30
(CM-343) Reorder default contact block

pezholio:
This adds functionality to "reorder" a default content block during the create/edit process. There has been a bit of faffing around required to enable us to render parts of a block, but this seems to work now, and should be generic enough to work with all types of composed block. For now I haven't added any Turbo functionality when clicking up / down to move a block, but this should be easy enough to add when we want to.

## Screenshots

image


app/components/edition/reorder_component.rb
```diff
@@ -0,0 +1,21 @@
...
+private
+
+  attr_reader :edition, :order, :redirect_path
+
+  def move_path(index, item, direction)
+    order_dup = order.dup
+    new_index = direction == :up ? index - 1 : index + 1
+    order_dup.insert(
+      new_index,
+      order_dup.delete_at(order_dup.index(item)),
```
graycodes, line 17: Sorry, you'll have to bear with me as I get used to rails. This line struck me as though it might be complex to delete at an index by getting something by looking up its index. I thought there might maybe be something available in the api like so?:
```ruby
order_dup.delete(item)
```

features/step_definitions/reorder_steps.rb
```diff
@@ -15,3 +15,10 @@
 
   assert_equal "Contact Form", items[0].find_all(".content-block__contact-key")[0].text
 end
+
+Then(/^I should see the contact form moved to the top$/) do
+  Capybara.current_session.driver.with_playwright_page do |page|
+    item = page.get_by_testid("reorder-item-0")
+    expect(item).to playwright_matchers.have_text("Contact Form")
+  end
+end
```
edavey, line 24: well done figuring out how to get the  assertions working with Turbo stream actions !

PR #85, opened 2025-10-01
(CM-509) Fix blocks ordered incorrectly on homepage

pezholio:
https://gov-uk.atlassian.net/jira/software/c/projects/CM/boards/134?selectedIssue=CM-509

Before, we had some text that claimed blocks were ordered by most recently updated first, but this wasn't the case. This should address the problem by calling the update call correctly, and removing some erroneous joins.


graycodes (approved):
Looks good to me!

PR #90, opened 2025-10-02
(CM-343) Ensure all items are rendered when reordering

pezholio:
If a block is created/edited, reordered, and then has a subsequent item added to it, when the user goes to reorder the items, the new item is missing, because it is not included in the order. This doesn’t matter when rendering the block, as any missing items are appended to the end, but this is not done when rendering the reorder component.

This fixes the issue by finding any missing items from the default order, and then adding these to the end of the order array when rendering the items in the reorder component.

### DIFF
--- a/app/components/edition/reorder_component.rb
+++ b/app/components/edition/reorder_component.rb
@@ -9,6 +9,11 @@ private
 
   attr_reader :edition, :order, :redirect_path
 
+  def item_keys
+    items_missing_from_order = edition.default_order - order
+    order + items_missing_from_order
+  end
+
   def move_path(position, item, direction)
     updated_order = order.dup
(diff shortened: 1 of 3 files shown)


app/components/edition/reorder_component.rb
```diff
@@ -9,6 +9,10 @@ def initialize(edition:, order:, redirect_path:)
 
   attr_reader :edition, :order, :redirect_path
 
+  def item_keys
+    order + (edition.default_order - order)
```
edavey, line 13: would it perhaps be helpful to extract `(edition.default_order - order)` to a method with a friendly name to make it clear what's being done here and why? If I understand correctly, that might be something like `elements_added_during_reordering()`?

edavey (approved):
the plot thickens, somewhat!

PR #91, opened 2025-10-02
(CM-517) Hide E2E blocks from users

pezholio:


features/search_for_object.feature
```diff
@@ -124,3 +124,14 @@ Feature: Search for a content object
     When the block "example search title" has been updated
     And I visit the Content Block Manager home page
     Then the block "example search title" should appear as the first item in the list
+
+  Scenario: E2E documents are not shown
+    When the block "example search title" has hide_from_search_index set to true
+    And I visit the Content Block Manager home page
+    Then I should not see the content block with title "example search title" returned
+
+  Scenario: E2E documents are shown to E2E users
```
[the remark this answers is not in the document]
pezholio: Yeah, that's fair, I can tweak the step to log in as a second user, rather than just faking the existing user

docs/architecture/decisions/0005-hide-e2e-related-content-from-users.md
```diff
@@ -0,0 +1,66 @@
...
+
+We will need to add the `E2E_EMAILS` environment variable to the application via the GOV.UK Helm charts. These are not
+secrets as such, so can be added in plain text.
+
+Once this is done, there will still be some E2E blocks that do not have this boolean field set, so we will need to 
+manually update them. We will do this via a Rake task instead of on the console as this is safer and more testable. 
+We can run the task manually and then delete the code once it has been run in all environments.
+
+This also gives us the flexibility to run the Content Modelling E2E tests more often (such as on a schedule,
+or after a deployment to integration or staging), because the risk of adding noise to the application will be lowered.
```
edavey, line 66: Do we also need to think about periodically purging the integration and staging environment of these unwanted content blocks? Or are we happy to have an every increasing volume of cruft in the databases?

In either case is there a case to be made that it would be simpler overall to deal with teardown of these tests as part of running the test harness? e.g. if each test run sends a `DELETE to /ops/content-block/{blockId}`. The code invoked could have guards in place e.g.:

- return if env is production
- return if document creator is not the E2E test user

app/models/user.rb
```diff
@@ -5,6 +5,12 @@ class User < ApplicationRecord
 
   validates :name, presence: true
 
+  def is_e2e_user?
+    ENV.fetch("E2E_EMAILS", "")
```
edavey, line 9: would `E2E_USER_EMAILS` be a littler clearer?

db/migrate/20251002140125_add_hide_from_search_index_field_to_documents.rb
```diff
@@ -0,0 +1,5 @@
+class AddHideFromSearchIndexFieldToDocuments < ActiveRecord::Migration[8.0]
+  def change
+    add_column :documents, :hide_from_search_index, :boolean, default: false, null: false
```
edavey, line 3: is it most straightforward to think of this flag in terms of `hide_from_search_index` or as something like `testing_artefact`? Perhaps that depends whether / how we are planning to ultimately delete these content blocks?

app/models/document/document_filter.rb
```diff
@@ -73,6 +73,7 @@ def to_date
   def unpaginated_documents
     documents = Document
     documents = documents.where(block_type: Schema.valid_schemas)
+    documents = documents.where(hide_from_search_index: false) unless Current.user&.is_e2e_user?
```
edavey, line 76: If we don't currently have other scenarios for needing to hide blocks, is it clearer to have just one term in play. e.g. for our finder to say:

- "exclude documents made in testing unless I'm the tester" rather than
- "exclude documents marked as hidden unless I'm the tester"

edavey (approved):
👍 all looks correct, but:

1. I'm not convinced that hiding an ever-increasing number of test blocks is better than cleaning up after each test
2. I wonder if we need the term "hide from search index?" ? Or whether talking directly about something like "test artefact?" would be more straightforward?

docs/architecture/decisions/0005-hide-e2e-related-content-from-users.md
pezholio: The problem we have is that these blocks don't just exist in our system - as soon as a block is published it makes a footprint in Publishing API and Content Store, as well as (if used) in other publishing apps. We do have the ability to delete content blocks from the app and Publishing API, but we can only do this if they're not used anywhere else within publishing.

app/models/user.rb
pezholio: Good plan - I've updated the Helm charts too https://github.com/alphagov/govuk-helm-charts/pull/3592

edavey:
> * We will eventually have a mechanism back for deleting blocks on a weekly basis, we won't have a huge amount of test content clogging up the system; and

ah, good to know that the periodic purges will return!

PR #93, opened 2025-10-03
CM-398: Change btn text from "publish" to "create"

graycodes:
[Jira Ticket](https://gov-uk.atlassian.net/browse/CM-398)

# Description

- Changes the text to be "Create" when first creating a new block.
- When editing an existing block the text remains "Publish" as before.
- When scheduling the content, the text remains "Schedule" as before.

Note: I've put some of my thinking in each of the commit comments if you want to see the thought process.

# Testing Steps 
- Create any new content block, right through to the final step.
  - Observe that the text on the final step of the flow (on the green button) is "Create".
- Edit this block and continue through the flow to the final step.
  - Observe that the text on the submit button is "Publish".
- Create a new block and continue through the flow, choosing to **Schedule** the content to go-live in future.
- Continue to the end of the creation flow.
  - Observe that the text on the submit button at the end of the flow is "Schedule"


features/support/helpers.rb
```diff
@@ -74,6 +74,10 @@ def click_save_and_continue
   end
 end
 
+def submit
+  find("button.govuk-button:not(.govuk-button--secondary)").click
```
pezholio, line 78: This is a good plan - but I'm trying to lean away from using CSS classes when looking for buttons / links and leaning more towards data attributes, so if we give the button a `data-testid` of `submit-button` (or whatever), we can more reliably find the button, and not have to worry about changes in CSS or text breaking tests

features/step_definitions/form_step_steps.rb
```diff
@@ -20,3 +20,7 @@
     assert_text "Add #{::Regexp.last_match(1).humanize(capitalize: false)}"
   end
 end
+
+Then("I should see a button labelled {string}") do |label|
+  expect(page).to have_button(label)
+end
```
pezholio, line 26: There's a Rubocop linting failure here - the easiest thing do do is run `bundle exec rubocop -A` to autofix any issues. We should probably have this in a pre-commit hook

pezholio (commented):
LGTM - just one minor thing about the Cucumber tests. Happy to chat about it 👍

features/step_definitions/form_step_steps.rb
pezholio: D'oh, just seen you'd fixed it :)

app/components/shared/continue_or_cancel_button_group.html.erb
```diff
@@ -7,7 +7,7 @@
       form: form_id,
       data_attributes: {
         "testid": "submit-button"
-      }
+      },
```
pezholio, line 10: We tend to rebase any lint failures into their relevant commits - I find [`git absorb`](https://github.com/tummychow/git-absorb) really useful for this

PR #97, opened 2025-10-06
CM 507 configure visibility of embed codes

edavey:


app/models/schema.rb
```diff
@@ -6,6 +6,7 @@ class Schema
 
   CONFIG_PATH = Rails.root.join("config/content_block_manager.yml").to_s
 
+  EMBED_CODE_VISIBLE_PROPERTY_KEY = "embed_code_visible".freeze
```
pezholio, line 9: I think this belongs in the previous commit

pezholio: Or possibly the first two commits are in the wrong order?

pezholio (approved):
LGTM - just a minor thing with the first two commits being out of order

app/models/schema.rb
edavey: thks, not sure how that happened!

PR #98, opened 2025-10-06
(CM-520) Preview improvements

pezholio:
While working on some ideas I had around host document previews, I noticed that a couple of things weren't working correctly, firstly the styling wasn't always loading correctly for the previews, and secondly, there are two "frontend" apps im GOV.UK that render different types of content. In integration, staging and production, there is a mechanism to work this out, but in dev, we don't have this same luxury. This makes an additional call to publishing API in dev to get the `rendering_app` for a piece of content and make the request to the relevant app. At the moment, we need to run these apps seperately. We should probably look into making both frontends a dependency in govuk-docker, but we need https://github.com/alphagov/govuk-docker/pull/898 to be merged first


test/unit/app/services/generate_preview_html_test.rb
```diff
@@ -158,4 +158,49 @@ class GeneratePreviewHtmlTest < ActiveSupport::TestCase
...
+        "rendering_app" => rendering_app,
+      }
+    end
+
+    before do
+      Rails.env.stubs(:development?).returns(true)
+      Services.publishing_api.stubs(:get_content).with(host_content_id).returns(publishing_api_response)
+    end
+
+    it "makes a request to the correct rendering app" do
```
edavey, line 176: is it worth being more specific here e.g.

> makes a request to the rendering app used by the Publishing API

test/unit/app/services/generate_preview_html_test.rb
```diff
@@ -158,4 +158,49 @@ class GeneratePreviewHtmlTest < ActiveSupport::TestCase
...
+
+      GeneratePreviewHtml.new(
+        content_id: host_content_id,
+        edition: block_to_preview,
+        base_path: host_base_path,
+        locale: "en",
+      ).call
+    end
+
+    describe "when the rendering app is not present" do
```
edavey, line 187: If I understand what's happening correctly, would it be correct to be more precise here with:

> When the Publishing API does not report a rendering app

edavey (approved):
Nicely done!

PR #101, opened 2025-10-07
(CM-535) Only show preview button for composed blocks

pezholio:
We initially added a preview button only in the grouped objects workflow step, which only shows for composed blocks. Since then, we also added the button at the review stage, which shows for all block types.

As non-composed blocks aren’t embeddable as a single block, then showing a preview of the block is irrelevant, so let’s hide the link to the preview.

### DIFF
--- a/app/views/editions/workflow/review.html.erb
+++ b/app/views/editions/workflow/review.html.erb
@@ -7,8 +7,10 @@
   } %>
 <% end %>
 
-<% content_for :title_side do %>
-  <%= render "shared/preview_link", edition: @edition %>
+<% if @edition.schema.embeddable_as_block? %>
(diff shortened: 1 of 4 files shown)


app/views/editions/workflow/review.html.erb
```diff
@@ -7,8 +7,10 @@
   } %>
 <% end %>
 
-<% content_for :title_side do %>
-  <%= render "shared/preview_link", edition: @edition %>
+<% if @edition.schema.embeddable_as_block? %>
```
edavey, line 10: Are "composed block" and "embeddable as block" synonyms? If so, should we refactor the naming to have `schema.composed_block?`?

PR #110, opened 2025-10-08
(CM-528) Show translated address labels

graycodes:
# Description

As per [CM-528](https://gov-uk.atlassian.net/browse/CM-528) the `Address name` field was still showing as `Title` in some places. This attempts to fix that.

We had a method on the summary card component called `nested_item_title` which I've removed in favour of replicating some of the functionality that's in this `TranslationHelper` to work with titles and labels in the same way.

The actual fix here is really just a one-line change in this component.

# Testing Steps

- In your locally running CBM
- Go to the Edit screen of a contact with an address (e.g. `/editions/:id/workflow/group_contact_methods`)
  - Observe that the text on the name of the address is `Address Name` and not `Title`
- Go to the Edit screen of a contact with a Telephone contact that has multiple (nested) phone numbers
  - Observe that these blocks are labelled `Telephone number 1`, `Telephone number 2` etc and that this is not broken.

Example correctly rendering blocks:
Screenshot 2025-10-08 at 14 19 27
Screenshot 2025-10-08 at 14 19 20

[CM-528]: https://gov-uk.atlassian.net/browse/CM-528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/document/show/embedded_objects/blocks_component.rb
+++ b/app/components/document/show/embedded_objects/blocks_component.rb
@@ -32,7 +32,7 @@ private
   def attribute_rows(key_name = :key)
     first_class_items(items).map do |key, value|
       {
-        "#{key_name}": key_to_title(key, schema_name),
+        "#{key_name}": key_to_label(key, schema_name),
         value: content_for_row(key, value),
         data: data_attributes_for_row(key),
       }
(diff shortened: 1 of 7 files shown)


pezholio (approved):
Lovely stuff - LGTM 👍

PR #506, opened 2026-03-03
(CM-825) Add validation to time period

pezholio:
This PR improves how date and time fields are handled in the content block manager by removing the intermediate `DateTimeConverter` abstraction and instead parsing date/time values directly from form inputs using the `ProcessedParams` class.

I've also had to make some tweaks to the component itself to parse date/time attributes fetching the params, rather than parsing the date/time directly (which throws a server-side error if invalid values are provided).

This has required a little bit of a tweak to the `object_params` method (and the controllers that use it), so we no longer alter the params in place when using `ProcessedParams`


app/models/schema/field.rb
```diff
@@ -41,6 +41,10 @@ def type
       @type ||= properties.fetch("type", "").inquiry
     end
 
+    def format
+      @format ||= properties.fetch("format", "").inquiry
+    end
```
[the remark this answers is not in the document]
pezholio: Good call!

spec/components/edition/details/fields/date_time_component_spec.rb
```diff
@@ -31,34 +46,48 @@
...
-      expect(page).to have_css('*[name="edition[details][date_range][start(1i)]"]') # year
-      expect(page).to have_css('*[name="edition[details][date_range][start(2i)]"]') # month
-      expect(page).to have_css('*[name="edition[details][date_range][start(3i)]"]') # day
-      expect(page).to have_css('*[name="edition[details][date_range][start(4i)]"]') # hour
-      expect(page).to have_css('*[name="edition[details][date_range][start(5i)]"]') # minute
+      expect(page).to have_css('*[name="edition[details][date_range][start][date(1i)]"]') # year
+      expect(page).to have_css('*[name="edition[details][date_range][start][date(2i)]"]') # month
+      expect(page).to have_css('*[name="edition[details][date_range][start][date(3i)]"]') # day
+      expect(page).to have_css('*[name="edition[details][date_range][start][time(4i)]"]') # hour
+      expect(page).to have_css('*[name="edition[details][date_range][start][time(5i)]"]') # minute
```
graycodes: 🤔 Sending these like this means you're referencing 2 datetimes, one which only has a date part and one which just has a time part. For convenience I see that it could be helpful having these as separate variables but I wonder if we're not going to pass these in one datetime then we could just abandon the multiparameter format and just send them as:
```
edition[details][date_range][start][year], 
edition[details][date_range][start][month], 
edition[details][date_range][start][day]
```
etc

spec/unit/app/models/processed_params_spec.rb
```diff
@@ -6,6 +6,57 @@
...
+    let(:fields) do
+      [
+        build(:field, name: "published_at", type: "string", format: "date", show_field: nil, nested_fields: nil),
+      ]
+    end
+
+    let(:params) do
+      {
+        "block_type" => {
+          "published_at(1i)" => "2024",
```
leenagupte: Are the blocks really passed in like this with "1i" etc in the name?

graycodes (commented):
Using this in the UI and I think there could be some issues with what we're doing here:

1. It's now a requirement on this screen that the day/month are zero-prefixed. If you type a day or month like `3` it fails (but `03` succeeds). This differs from how it works on the schedule page where a day/month like `3` would succeed. 
2. When you type one valid date and one blank date it's parsing your inputs and re-populating the fields in an invalid format. For example if I type `03` `03` `2003` in the `start` date fields and some valid time. Then in the `end` range I leave this blank. When I submit the form it then re-populates my `start` date fields with `3` `3` `2003` (no leading zeroes) which is now an invalid format so I can't resubmit the form.
3. As it's no longer building a date, you can type things like `-3` in the `year` field and because of the split/join on the hyphen character, when I submit the form I end up with `3` in the `month` field (not year!).

app/models/processed_params.rb
```diff
@@ -29,6 +32,14 @@ def processed_fields(fields, collection)
     collection
   end
 
+  def process_date_field(field, collection)
+    (1..3).map { |i| collection.delete("#{field.name}(#{i}i)") }.join("-")
```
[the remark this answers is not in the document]
leenagupte: If the date always follows the format of `"{field.name}(1i)"` e.g. `"published_at(1i)"` and there are only ever three parts to a date. It'd be much more readable to put everything on a separate line, e.g. `year = "#{field.name}(1i)"`

and then return "#{year}-#{month}-#{day}"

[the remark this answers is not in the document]
leenagupte: The same can be applied to `process_time_field` which only has two parts.

leenagupte (commented):
Partial review of the first 6 commits. I'll try and get to the rest later on today.

I have few inline comments.

app/models/processed_params.rb
```diff
@@ -29,6 +32,21 @@ def processed_fields(fields, collection)
     collection
   end
 
+  def process_date_field(field, collection)
+    year = collection.delete("#{field.name}(1i)")
```
leenagupte: You don't need to delete here.

spec/unit/app/models/processed_params_spec.rb
```diff
@@ -6,6 +6,96 @@
...
+            "published_at(2i)" => "1",
+            "published_at(3i)" => "1",
+          },
+        }
+      end
+
+      it "pads the date parts with zeroes" do
+        expect(processed_params.result).to eq({
+          "block_type" => {
+            "published_at" => "2024-01-01",
```
leenagupte: It's great that you've added this check for the `rjust`

graycodes (changes requested):
We mobbed on some testing of the validation here and had a concern about a particular case. When the user inputs something like 
`29-2-2024`
this is a valid correct date and is accepted, as 2024 was a leap year.

However when a user inputs
`29-2-2025`
It is not accepted. This in itself is fine as 2025 is not a leap year. However, the UI then shows me a page with my date inputs highlighted in red, pre-filled with the date
`1-3-2025`
This appears to be because when you parse the above invalid date, ruby decides that the 29th of Feb on a non-leap year is *really* just another way of referring to the 1st March. 

I can see why it does this but we probably shouldn't show the user this 1st March date.

app/models/edition/date_time_parts.rb
```diff
@@ -0,0 +1,59 @@
...
+    def param_value(key)
+      params.dig(:edition, :details, block_type, field_name, key)
+    end
+
+    def parse_time_from_details
+      date = edition.details.dig(block_type, field_name, date_field.name) || ""
+      time = edition.details.dig(block_type, field_name, time_field.name) || ""
+      Time.zone.iso8601("#{date}T#{time}")
+    rescue ArgumentError
+      nil
```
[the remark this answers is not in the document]
leenagupte: Will this be a replacement for `format_date_minimum` or is that being replaced?

leenagupte (commented):
I've gotten as far as reviewing up to commit [Add a DateTimeParts class](https://github.com/alphagov/content-block-manager/pull/506/changes/f26f29c0a53f92b792622ede72c6a98501a23bff) and had to stop.

I'm new to this code base, so reviewing was always going to be a challenge. I think that could be helped by restructuring the commits so that follow on from each other. One commit builds on another and so forth to tell a story.

I am quite confused by what is happening here. The title of the PR is to add validation for a time period, but that seems quite hidden away, and not easy to find at all.

I think it is a good idea to use the Ruby Time helpers to validate if a time period contains validate dates and times, but think it should happen far earlier in the save flow.

I can pick up the review again next week.

spec/unit/app/models/processed_params_spec.rb
leenagupte: Can a test be added for the date fields not having a value, and a non-numeric value, so that we know what the expected behaviour is.

app/models/processed_params.rb
```diff
@@ -45,4 +63,9 @@ def update_conditional_object(field, collection)
   def cast_to_boolean(value)
     ActiveRecord::Type::Boolean.new.cast(value) || false
   end
+
+  def get_and_remove_date_time_part_from_collection(field, part, collection)
+    index = %w[year month day hour minute].index(part) + 1
+    collection.delete("#{field.name}(#{index}i)") || ""
```
leenagupte: So the params object is being modified here? Is that on purpose? Can you add a note to the commit message saying why that's needed?

leenagupte (commented):
Sorry for the delay, I've finished reviewing all of the commits. 

I've some inline questions and comments, I've also unresolved some of my previous comments where I couldn't see where/how they were resolved.

Could you respond the comments in the comment thread so I can see your responses rather than just resolving them, that way I'll get a handy notification of the change.

leenagupte (commented):
Thanks for making changes from last time.

I do feel like this is still quite overly complicated, even the test data feels that way.

I've left some inline comments. The main point of contention is that a lot of hoops are being jumped through because the user input is being modified before it is validated. I very much feel that this is the wrong order and has made the code even more complex than it needs to be. 
I accept that this code will be complex, so any opportunities to simplify should be taken.

I think perhaps when validation is handled needs to be revisited. I'd be happy to join in any conversations about that.

pezholio:
Closing this, as it's now been addressed in https://github.com/alphagov/content-block-manager/pull/581

PR #553, opened 2026-03-30
Experimental: Use ActiveRecord models rather than JSONSchema to model block types

edavey:
See the ADR (which is the first commit) for an overview of this experiment:

---

# 11. Experiment with Traditional ActiveRecord Models for Content Blocks

Date: 2026-03-11

Status: Experimental

### Current State

The Content Block Manager currently uses a **schema-driven generic approach** for modeling content blocks:

1. **Single Document Model:** All block types use a single `Document` model (no subclasses).
2. **Single Edition Model:** All block types use a single `Edition` model with a generic `details` JSON column.
3. **JSON Schema Validation:** Block type structure is defined and validated via JSON schemas stored in `app/models/schema/definitions/*.json`.
4. **Generic Forms:** Form interfaces are dynamically generated from the JSON schemas.

This architecture provides flexibility and allows new block types to be added by defining a JSON schema without creating new models or migrations.

This was a necessary technical design initially as all schemas were defined in Publishing API. However, in order to have greater control, particularly over validation, we've recently moved the schema into this repository (See [ADR 10 Make Content Block Manager the source of truth for schemas][])

The use of JSON schema definition is no longer a requirement.

### The Question

As the application grows, we're evaluating whether the schema-driven approach provides the best developer experience and maintainability, or whether a **traditional Rails ActiveRecord approach** with dedicated models and tables might be clearer and easier to work with.

Specifically, we want to understand:

2. **Developer Experience:** Is it easier and faster to work with explicit Rails models and associations vs. generic JSON columns? Our experience when adding a new block type is that there are numerous differences or additional features which are needed. At face value they appear minor but in reality prove complex to introduce to an application model which is entirely generic. Our small number of classes can do everything, but at what price in terms of complexity and cognitive load.
3. **Maintainability:** Are dedicated models with typed columns easier to understand and modify than schema definitions?
4. **Adding Fields:** Is it simpler to add a migration for a new column vs. updating a JSON schema?
5. **Performance:** Do typed columns with proper indexes perform better than querying within JSON?
6. **Testing:** Are traditional model tests more straightforward than testing schema validation?

## Decision

We will conduct an **experiment** by implementing an alternative architecture using traditional ActiveRecord models, running **in parallel** with the existing schema-driven system.

### Experiment Scope

We will implement the **TimePeriod** block type as a proof-of-concept using the new architecture. This will allow us to:

- Compare both approaches side-by-side in the same codebase
- Evaluate real-world developer experience
- Make an informed decision about which direction to pursue

This is **not yet** a decision to migrate away from the existing schema-based paradigm - it is purely exploratory.

---

## Current state

current_state

## Proposed target state

proposed_target_state

---

To try this out locally you can either:

## 1. Use some conditionally applied UI elements

See some conditional UI elements by giving yourself the `schemaless_experiment` permission:

```rb
User.last.permissions << `schemaless_experiment`
User.last.save
``` 

or by adding `?schemaless_experiment=true` to your URL

## 2. Navigate manually to the new-style TimePeriod form

`/block/time_period_editions/new`

new_time_period_edition_form

## Screenshot of new style TimePeriod

new_style_time_period

### DIFF
--- a/app/models/block/time_period_date_range.rb
+++ b/app/models/block/time_period_date_range.rb
@@ -1,0 +1,34 @@
+module Block
+  class TimePeriodDateRange < ApplicationRecord
+    self.table_name = "block_time_period_date_ranges"
+
+    belongs_to :edition, class_name: "Block::TimePeriodEdition"
+
+    validates :start, presence: true
+    validates :end, presence: true
+    validate :end_date_after_start_date
+
+    def to_details
+      {
+        "start" => {
+          "date" => start.strftime("%Y-%m-%d"),
+          "time" => start.strftime("%H:%M"),
+        },
+        "end" => {
+          "date" => self.end.strftime("%Y-%m-%d"),
+          "time" => self.end.strftime("%H:%M"),
+        },
+      }
+    end
+
+  private
+
+    def end_date_after_start_date
+      return if self.end.blank? || start.blank?
+
+      if self.end <= start
+        errors.add(:end, "must be after start date")
+      end
+    end
+  end
+end
--- a/app/models/block/time_period_date_range.rb
+++ b/app/models/block/time_period_date_range.rb
@@ -2,6 +2,9 @@ module Block
   class TimePeriodDateRange < ApplicationRecord
     self.table_name = "block_time_period_date_ranges"
 
+    include DateValidation
+    date_attributes :start, :end
+
     belongs_to :edition, class_name: "Block::TimePeriodEdition"
 
     validates :start, presence: true
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_03_11_174532) do
+ActiveRecord::Schema[8.1].define(version: 2026_03_11_175854) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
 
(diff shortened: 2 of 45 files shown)


pezholio:
More thoughts! I think I'd be happy giving this a go, but I'd rather the unit of difference was just the model, rather than having separate views and controllers etc. With that, we can define validations and data manipulation in the model, as well as potentially the workflow shape too.

app/models/block/time_period_date_range.rb
```diff
@@ -0,0 +1,34 @@
...
+  class TimePeriodDateRange < ApplicationRecord
+    self.table_name = "block_time_period_date_ranges"
+
+    belongs_to :edition, class_name: "Block::TimePeriodEdition"
+
+    validates :start, presence: true
+    validates :end, presence: true
+    validate :end_date_after_start_date
+
+    def to_details
```
pezholio, line 11: I wonder if we should be consistent with other publishing apps and use presenters to present information we send to the Publishing API - example here https://github.com/alphagov/whitehall/blob/main/app/presenters/publishing_api/case_study_presenter.rb

db/schema.rb
```diff
@@ -34,7 +34,7 @@
     t.datetime "created_at", null: false
     t.text "description"
     t.text "instructions_to_publishers"
-    t.integer "lead_organisation_id"
+    t.uuid "lead_organisation_id"
```
graycodes, line 37: Hmm I would have thought this would be the only change in this file. Are the rest of these changes valid?

graycodes (commented):
Wonderful work!

It's taken me a while to get around to reviewing this in detail but I've finally had a thorough look through and I still feel like this is the right direction. You tackled some parts I hadn't considered too. 

Looking at this now it feels already like these commits are very easy to follow, and avoid a lot of the complexity that we have had to deal with re schemas and validation.

I feel like there may be some commonalities that arise once we start implementing multiple models in this format, but we can worry about those later.

Looking forward to progressing this soon.

PR #580, opened 2026-04-08
(CM-904) Show diff of content in previews

pezholio:
This improves the previews to show a diff of the content where blocks are used, rather than just highlighting the blocks in question. 

~~NOTE: At the moment, diffs within headings aren't rendering correctly. Need to fix this upstream, and also check with Mainstream if any proposed fix messes with any of their use cases.~~

## Screenshot

2026-04-09 14 44 01 content-block-manager integration publishing service gov uk b62c66cac31c

### DIFF
--- a/engines/block_preview/app/model/block_preview/preview_html.rb
+++ b/engines/block_preview/app/model/block_preview/preview_html.rb
@@ -23,7 +23,8 @@ module BlockPreview
       add_draft_style(nokogiri_html)
       update_css_hrefs(nokogiri_html)
       update_js_srcs(nokogiri_html)
-      nokogiri_html.at_css("[data-module=\"govspeak\"]").replace BlockPreview::ContentDiff.new(nokogiri_html, block).to_s
+      add_nokodiff_stylesheet(nokogiri_html)
+      update_preview_with_diff(nokogiri_html)
       update_local_link_paths(nokogiri_html)
       update_local_form_actions(nokogiri_html, uri.scheme, uri.host)
       nokogiri_html.to_s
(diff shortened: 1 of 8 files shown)


engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -18,19 +20,17 @@ def initialize(content_id:, block:, base_path:, locale:, state:, auth_bypass_id:
     def to_s
       uri = Addressable::URI.parse(frontend_path)
       nokogiri_html = html_snapshot_from_frontend(uri)
-      update_local_link_paths(nokogiri_html)
-      update_local_form_actions(nokogiri_html, uri.scheme, uri.host)
       add_draft_style(nokogiri_html)
       update_css_hrefs(nokogiri_html)
       update_js_srcs(nokogiri_html)
-      replace_existing_content_blocks(nokogiri_html).to_s
+      nokogiri_html.at_css("[data-module=\"govspeak\"]").replace BlockPreview::ContentDiff.new(nokogiri_html, block).to_s
```
edavey: would it be more consistent to use a method which take and returns `nokogiri_html`? e.g. 

```
def insert_diff(nokogiri_html)
  nokogiri_html.at_css("[data-module=\"govspeak\"]")
  .replace(
    BlockPreview::ContentDiff.new(nokogiri_html, block).to_s
  )

  nokogiri_html
end
```

That seems to be the pattern here? And might keep `to_s` reading as a more declarative list?

edavey (approved):

engines/block_preview/app/model/block_preview/preview_html.rb
pezholio: Good plan - have fixed!

PR #581, opened 2026-04-08
CM 825: validate time period date range

edavey:
See Jira [CM-825: Validate TimePeriod#date_range](https://gov-uk.atlassian.net/browse/CM-825)

---

In this PR we:

### 1. Change the format of TimePeriod#date_range

We're changing TimePeriod.date_range to be:
```
- start: datetime
- end: datetime
```

rather than:

```
- start:
  - date: date
  - time: time

- end:
  - date: date
  - time: time
```
This will make the work of validating user input a little simpler, in
that for:

- validation
- serialision/deserialision

we'll be acting on a pair of datetimes, without the additional step of
decomposing to date and time values, which have no useful function on
their own.

### 2. Add validation

We now:

- validate that dates are real
- validate before persisting any values
- re-render the exact form contents when there's a validation error

## Plan to deploy these changes

- ✅  extend "tools" to support new TimePeriod#date_range structure whilst
  remain backwards compatible with the original structure (see https://github.com/alphagov/govuk_content_block_tools/pull/143)

- release data migration on Content Block Manager to convert existing
  TimePeriods to new structure (**this PR**)

- change Content Block Manager to create and edit Time Periods with new
  structure (**this PR**)

- remove support for the old Time Period structure from "tools" and remove the temporary `data_migration:convert_time_period_date_range_to_datetime` task from this repo


app/validators/details_validator/custom_validators.rb
```diff
@@ -2,11 +2,11 @@ class DetailsValidator < ActiveModel::Validator
...
     def format_date_minimum(body)
       proc do |instance, schema, _instance_location|
-        return true unless instance.is_a?(String)
-        return true unless schema["format"] == "date"
+        next true unless instance.is_a?(String)
+        next true unless schema["format"] == "date"
 
         min_str = schema["formatMinimum"]
-        return true if min_str.nil?
+        next true if min_str.nil?
```
pezholio, line 9: Nice, I noticed this when I started looking at this. D'oh!

app/validators/details_validator/custom_validators.rb
```diff
@@ -3,22 +3,27 @@ module CustomValidators
...
         next true unless instance.is_a?(String)
-        next true unless schema["format"] == "date"
 
+        format = schema["format"]
         min_str = schema["formatMinimum"]
         next true if min_str.nil?
 
+        unless format == "date-time"
+          raise ArgumentError, "formatMinimum is only supported for date-time fields, got format: #{format.inspect}"
+        end
```
pezholio, line 13: Ah, smart!

app/models/schema/field.rb
```diff
@@ -158,6 +162,18 @@ def permitted_params
 
   private
 
+    def raise_nested_field_not_supported_error_for(nested_field_name)
```
pezholio, line 161: Like this!

spec/unit/app/models/schema/field_spec.rb
```diff
@@ -326,6 +326,37 @@
...
+              "type" => "string",
+            },
+          },
+        }
+      end
+
+      let(:field) { Schema::Field.new("start", Schema.new("test", body)) }
+
+      let(:expected_error_message) do
+        %~
```
pezholio, line 344: I didn't know you could do `%~` for HEREDOCs, much easier syntax to remember!

pezholio (commented):
Couple of drive-by comments, will do a full review when ready

app/public/domain/date_time_converter.rb
```diff
@@ -1,4 +1,50 @@
 class DateTimeConverter
+  attr_reader :date_time, :errors, :raw_params
+
+  FieldData = Data.define(:year, :month, :day, :hour, :minute,
```
pezholio: This is quite a big class now - is it worth putting it in its own file? (possibly namespaced under `date_time_converter`)

app/models/schema/definitions/time_period.json
```diff
@@ -8,37 +8,18 @@
...
-              "format": "date"
-            },
-            "time": {
-              "type": "string",
-              "format": "time"
-            }
-          }
+          "type": "string",
+          "format": "date-time",
+          "x-custom-format": "datetime"
```
pezholio: Do we need `x-custom-format` now?

app/models/schema/field.rb
```diff
@@ -151,8 +151,12 @@ def permitted_params
         (1..3).map { |i| "(#{i}i)" }
       elsif format.time?
         (4..5).map { |i| "(#{i}i)" }
-      elsif properties["x-custom-format"] == "datetime"
+      elsif properties["x-custom-format"] == "datetime" && nested_fields.present?
```
pezholio: We can probably use `properties["format"]` here

app/controllers/editions/embedded_object_controller.rb
```diff
@@ -74,6 +71,12 @@ def set_schema_and_subschema
     )
   end
 
+  def build_object_from_params(details)
+    raise ArgumentError, "Only date_range is supported" unless details["date_range"]
+
+    validate_and_convert_date_range(details["date_range"])
+  end
```
pezholio: Will this not throw an error if `date_range` is not present? (i.e. all other block types?) I imagine no, as the tests are passing, but I'm confused why it doesn't error

app/components/edition/details/fields/date_time_component.rb
```diff
@@ -1,4 +1,6 @@
 class Edition::Details::Fields::DateTimeComponent < ViewComponent::Base
+  RawDateTime = Data.define(:year, :month, :day, :hour, :min)
```
pezholio: Wonder whether this (and the associated methods) should be a class on their own?

pezholio (commented):
Couple more suggestions / comments. Looks great overall though!

app/public/domain/date_time_converter.rb
edavey: 👍 will look into this

app/models/schema/definitions/time_period.json
edavey: hmm, will investigate, thks!

app/controllers/editions/embedded_object_controller.rb
edavey: hmm, I'll clarify that...

app/components/edition/details/fields/date_time_component.rb
edavey: thks, will take a look

edavey:
@pezholio I've made some changes to the `EmbeddedObjectController` so that it no longer has knowledge of particular subschemas (`date_range`). Let me know if you think it's too magical!

app/controllers/editions/embedded_object_controller.rb
```diff
@@ -73,10 +72,21 @@ def set_schema_and_subschema
...
   end
 
-  def build_object_from_params(details)
-    raise ArgumentError, "Only date_range is supported" unless details["date_range"]
+  def validate_and_convert_object(object_params)
+    validator_module = validator_for_subschema
+    return object_params unless validator_module
 
-    validate_and_convert_date_range(details["date_range"])
+    include_validator(validator_module)
```
pezholio: If we're getting the module name, would it not make more sense for the validators to be classes that we call directly, rather than including them?

app/public/domain/date_time_converter.rb
edavey: done

app/models/schema/definitions/time_period.json
edavey: yes, that's tidied up now, thanks

app/models/schema/field.rb
edavey: yes, I've tidied this up now, thks

app/controllers/editions/embedded_object_controller.rb
edavey: This has now been reworked to be more generic and no longer have knowledge of a particular subschema (`date_range`)

app/components/edition/details/fields/date_time_component.rb
edavey: done

app/controllers/editions/embedded_object_controller.rb
edavey: 👍 this is now a class, which make it a bit less odd and magical...

app/components/edition/details/fields/date_time_component.rb
```diff
@@ -1,4 +1,7 @@
 class Edition::Details::Fields::DateTimeComponent < ViewComponent::Base
+  MULTIPARAMETER_INDEX = { year: 1, month: 2, day: 3, hour: 4, minute: 5 }.freeze
```
pezholio, line 2: Nice!

app/components/edition/details/fields/date_time_component.rb
```diff
@@ -1,4 +1,7 @@
...
+  LABELS = { year: "Year", month: "Month", day: "Day", hour: "Hour", minute: "Minute" }.freeze
```
pezholio: Would we be better using translations here?

app/models/schema/field.rb
```diff
@@ -151,8 +151,12 @@ def permitted_params
...
+        # Old format: datetime object with nested date/time fields, to be removed
```
pezholio: Maybe make this comment into a `TODO` format - to make it easier to find when tidying up?

lib/tasks/convert_time_period_date_range_to_datetime.rake
```diff
@@ -0,0 +1,45 @@
+namespace :data_migration do
+  desc "Convert TimePeriod#date_range to datetime format"
+  task convert_time_period_date_range_to_datetime: :environment do
+    log = proc { |str|
+      timestamp = Time.current.strftime("%F %T")
+      line = "#{timestamp} > #{str}"
+      puts line unless Rails.env.test?
+      Rails.logger.info(line)
+    }
```
pezholio, line 9: Clever!

app/controllers/editions/embedded_object_controller.rb
pezholio: Sweet!

pezholio (approved):
LGTM - thanks for all the hard work on this!

app/models/schema/field.rb
edavey: this is actually removed within this branch

app/components/edition/details/fields/date_time_component.rb
edavey: good point

edavey: done

app/models/schema/field.rb
pezholio: Ah yeah, of course!

app/public/domain/date_and_time/field_data.rb
```diff
@@ -0,0 +1,45 @@
...
+      hour = params["#{field_name}(4i)"]
+      minute = params["#{field_name}(5i)"]
+
+      new(
+        year:,
+        month:,
+        day:,
+        hour:,
+        minute:,
+        year_int: parse_integer(year),
```
graycodes: I wonder if there is a potential issue with this class where we've defined it as something which has both string and int versions of its fields available to be passed in. I imagine that it's not likely we'll ever want to pass in both sets of arguments, so would it make sense for one to be defined by the other? e.g. We would not want someone to do this:
```ruby
f = FieldData.new(
  year: "2002", ... 
  year_int: 2001, ...
)
```

So would it make sense for us to always pass in one set (the params?) and have the others defined by methods or something?

(Of course, I've not seen how this is going to be used yet so I might be wrong)

app/public/domain/date_and_time/converter.rb
```diff
@@ -0,0 +1,67 @@
...
+  private
+
+    def parse_from_params(params, field_name)
+      fields = DateAndTime::FieldData.from_params(params, field_name)
+
+      return add_error(:date_blank) if fields.all_date_fields_blank?
+      return add_error(:date_invalid) if fields.any_date_field_unparseable?
+      return add_error(:date_invalid) if fields.negative_day_or_month_provided?
+
+      ensure_date_is_valid(fields)
```
graycodes: v minor: I assume the ArgumentError is here to catch when the Date.new call fails. For clarity, is it worth wrapping this `ensure_date_is_valid` call in a begin?

app/public/domain/date_and_time/field_data.rb
edavey: yes, there is definitely something icky here... It started off as an internal data class but now that it's been extracted, the big constructor signature is confusing and, as you point out, dangerous...

I'll take another look

edavey: I've refactored this. It's now a regular class with the construction signature of 

```rb
def initialize(year:, month:, day:, hour: nil, minute: nil)
```

Setting up the "integer" versions of the these values is handled internally now. I've also renamed the class to `FieldValues` which I think is more accurate.

app/public/domain/date_and_time/converter.rb
edavey: hmm, thanks for pointing out that this is a little unclear/indirect. As the use of "implicit rescue" is generally felt to be desirable for its expressiveness I've reworked this to make it clearer exactly what's behind that `ArgumentError`. We now have an `UnparseableDateError`:

```rb
    def parse_from_params(params, field_name)
      values = DateAndTime::FieldValues.from_params(params, field_name)

      return add_error(:date_blank) if values.all_date_fields_blank?
      return add_error(:date_invalid) if values.any_date_field_unparseable?
      return add_error(:date_invalid) if values.negative_day_or_month_provided?

      ensure_date_is_valid(values)
      @date_time = build_datetime(values)
    rescue UnparseableDateError
      add_error(:date_invalid)
    end

    def ensure_date_is_valid(values)
      Date.new(values.year_int, values.month_int, values.day_int)
    rescue ArgumentError
      raise UnparseableDateError
    end
```

Does this work better?

app/models/schema/field.rb
```diff
@@ -147,15 +147,7 @@ def permitted_params
...
-      elsif format.date?
-        (1..3).map { |i| "(#{i}i)" }
-      elsif format.time?
-        (4..5).map { |i| "(#{i}i)" }
-      elsif properties["x-custom-format"] == "datetime" && nested_fields.present?
-        # Old format: datetime object with nested date/time fields, to be removed
-        # once dual-running is no longer needed
-        nested_fields.flat_map(&:permitted_params).map { |field| "#{name}#{field}" }
-      elsif properties["x-custom-format"] == "datetime"
+      elsif datetime_format?
```
graycodes, line 150: 🙌🏻

graycodes (approved):

PR #600, opened 2026-04-15
(CM-914) Update Ruby version to Ruby 4

pezholio:
While current Ruby versions (3.3 and 3.4) are still under normal maintenance and not yet at End-of-Life (EOL), Ruby 3.3 is expected to reach EOL on 31 March 2027. Planning this upgrade proactively allows us to avoid last-minute migrations, reduce risk. Moving early to Ruby 4.0 ensures we stay ahead of the support lifecycle, benefit from performance and language improvements, and maintain a secure, consistent runtime across all applications.


graycodes (commented):
Do we need to now provide these binstubs for this to work once deployed?

https://gds.slack.com/archives/CAB4Q3QBW/p1775749508903889

pezholio:
We have all the binstubs we need already in `bin` - I might deploy this to integration first to confirm that though!

pezholio:
Added a few more binstubs, and deployed to integration - all seems to be working!

patrickjfl (approved):
Looks good as far as I can see, nice work getting ahead of the game Stu.

graycodes (approved):

PR #603, opened 2026-04-16
(CM-816) Resolve packwerk todo tasks

graycodes:
# Jira Ticket
https://gov-uk.atlassian.net/browse/CM-816

# Description

We left a few files in the packwerk `package_todo.yml` file as part of a previous PR. These files were test files which referenced models in the main app directly, which breaks packwerk's rules.

In this PR I have simply moved these models into the public dir. They are not Rails ActiveRecord models, simply plain ruby classes so this should be safe enough to do.

I think philosophically it's the right thing to do to share these models across both the main app and the engines too, as they're needed to render the diffs of the block changes.


pezholio (approved):
Yep, happy with this 👍 Thanks for looking into it!

PR #610, opened 2026-04-20
(CM-838) Prevent duplicate names for content blocks and block components

pezholio:
Currently, CBM allows usage of the same name for a content block. This can cause problems when assisting users to find the correct block.

This adds validations to show an error when a user attempts to create a block with the same name as an existing block. There may be a legitimate reason for the block's name, so we allow the user to skip this validation when submitting again (inspired by Whitehall's approach to the same problem)

## Screenshot

image


features/title_uniqueness.feature
```diff
@@ -0,0 +1,31 @@
...
+  Scenario Outline: Creating a block with a non-unique title
+    When I visit the Content Block Manager home page
+    And I click to create an object
+    And I click on the "<schema>" schema
+    And I complete the form with the following fields:
+      | title     | description      | organisation        |
+      | my title  | Some description | Ministry of Example |
+    Then I should see an error message indicating that the title is not unique
+    When I save and continue
+    Then I should not see the error message indicating that the title is not unique
```
edavey: Perhaps something like:

> Then the duplicate title validation check will have been skipped

is clearer?

spec/integration/editions_controller_spec.rb
```diff
@@ -83,6 +83,38 @@
...
+        end
+
+        it "redirects to the review links step when ignore_title_uniqueness_check is set" do
+          redirects_to_step(:review_links) do
+            post document_editions_path(new_document), params: {
+              "edition": {
+                document_attributes:,
+                details:,
+                lead_organisation_id: organisation.id,
+                ignore_title_uniqueness_check: "1",
```
edavey: Could/should we make it clearer that `ignore_title_uniqueness_check` is used to record the fact that the user is accepting the risks associated with re-using a block title which already exisits?

e.g. something like `accept_risk_of_duplicate_title` ?

config/locales/en.yml
```diff
@@ -10,6 +10,8 @@ en:
               contact_methods:
                 blank: Select a contact method
         edition:
+          title:
+            not_unique: "A content block with the same title already exists. If you continue, there will be duplicate blocks with the same name."
```
edavey: a bit nit-picky, but I'm wondering if this is a bit inaccurate? I think the point is that the user is on the verge of creating a different block with a duplicate title (rather than a duplicate block)? i.e. that our warning should actually be something along the lines of:

> A content block with the same title already exists. If you continue, this block will not have a unique title.

features/title_uniqueness.feature
```diff
@@ -0,0 +1,31 @@
...
+
+  Examples:
+    | schema      |
+    | pension     |
+    | contact     |
+    | time_period |
+    | tax         |
```
edavey, line 31: this is neat!

edavey (approved):
👍 I've made a few minor comments/suggestions.

spec/integration/editions_controller_spec.rb
pezholio: I like that!

features/title_uniqueness.feature
pezholio: Good call 👍

config/locales/en.yml
pezholio: Yeah, that's fair. Will fix 👍

PR #611, opened 2026-04-20
(CM-925) Add a cooldown period to Dependabot

pezholio:
In response to recent ecosystem attacks on dependencies, we need to add a cooldown period to Dependabot, so Dependabot waits a few days after a version is published before raising a PR. This gives the community time to flag compromised or broken releases.

I’ve added an exclusion for GDS-managed gems, as we have more control over releases.

This does not affect critical security patches. These will still be raised by Dependabot without delay.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,13 +5,23 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+  cooldown:
+    default-days: 3
+    exclude: # Exclude GDS-managed gems from the cooldown period
+      - govuk_*
+      - content_block_tools
+      - nokodiff
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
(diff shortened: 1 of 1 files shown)


graycodes (approved):
I think this is a good idea, my only thought is that I think we're all looking at doing the same thing in different teams, but picking different numbers:

content-data-api (10 days)
https://github.com/alphagov/content-data-api/pull/2496/changes

gds-api-adapters (7 days)
https://github.com/alphagov/gds-api-adapters/pull/1390/changes

(and here, 3 days)

But I guess the important part is that it's set to something sensible.

pezholio:
The guidance (coming soon) is a _minimum_ of 3 days - `gds-api-adapters` seems to go for 3 too?

graycodes:
Ah yes sorry, it's their "interval" that's weekly

PR #613, opened 2026-04-21
(CM-913) Make outcomes into single step processes 

graycodes:
# Jira Ticket
https://gov-uk.atlassian.net/browse/CM-913

# Description
1. Condenses the outcome process down into a one-step process, rather than the current 2 page flow.
2. Allows the performer name to be optional.

# Notes
You can see that on the timeline items we show the same description for the state transition as for the outcome event. This can/will be tidied up in the next task to allow the fact check/review processes to be completed in any order (https://gov-uk.atlassian.net/browse/CM-921).

# Screenshots

https://github.com/user-attachments/assets/3ac54b20-4abb-4d18-a680-add7baa1670f


config/locales/en.yml
```diff
@@ -79,11 +79,12 @@ en:
     body:
       edition:
         fact_check:
-          performed: "Fact check performed by %{performer}"
+          performed: Fact check performed
```
pezholio: I feel like having two keys here, `performed` and `performed_with_performer` (or similar) might be better here:

```
fact_check:
  performed: "Fact check performed"
  performed_with_performer: "Fact check performed by %{performer}"
```

Then the code in `domain_event_outcome_resolution` would be a bit simpler.

pezholio (commented):
Looks good, just one point about the translations, which I think will tighten things up

app/components/document/show/document_timeline/timeline_item_component.rb
```diff
@@ -83,7 +83,11 @@ def domain_event_outcome
   end
 
   def domain_event_outcome_resolution
-    I18n.t("domain_event.body.#{domain_event.name}", performer: domain_event.metadata["performer"])
+    if domain_event.metadata["performer"].present?
+      I18n.t("domain_event.body.#{domain_event.name}_with_performer", performer: domain_event.metadata["performer"])
+    else
+      I18n.t("domain_event.body.#{domain_event.name}")
+    end
```
pezholio, line 90: 🙌

pezholio (approved):
LGTM 👍

edavey (approved):
Very clear, thks. 

It might be helpful if we add a link to the Jira task you mention:

> next task to allow the fact check/review processes to be completed in any order.

assuming that already exists?

PR #615, opened 2026-04-22
(CM-928) Move configuration to schema

pezholio:
A while ago, we moved to the use of local schemas. Part of the rationale for this was so we could have one source of truth for configuration and data shape. We haven’t done this yet, and as we are starting to consider how we move forward with schemas and configuration, moving to a single source will make migration easier.


edavey (approved):
I can see a minor benefit in consolidating all the interlocking schema/config in one place.

And I appreciate that this is an initial step in a larger refactoring plan. It might be useful to sketch out that plan in at least broad strokes? Will it perhaps lead to a world without any JSONSchema definitions?

PR #620, opened 2026-04-28
(CM-931) Show available formats on edition page

edavey:
See  Jira [CM-931](https://gov-uk.atlassian.net/browse/CM-931)

---

## User story

- So that I can copy the embed code for the format which I want to use
- As an editor viewing a particular block
- I want to see its various formats as well as the “default block”

---

In this PR we use the new format "specifier functionality" introduced in Content Block Tools [PR 147](https://github.com/alphagov/govuk_content_block_tools/pull/147) to show all available formats on the main "edition page" after the "default block".

available_formats

[CM-931]: https://gov-uk.atlassian.net/browse/CM-931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/edition/show/format_block_component.rb
+++ b/app/components/edition/show/format_block_component.rb
@@ -1,0 +1,49 @@
+class Edition::Show::FormatBlockComponent < ViewComponent::Base
+  def initialize(edition:, format:)
+    @edition = edition
+    @document = edition.document
+    @format = format
+  end
+
+private
+
+  attr_reader :edition, :document, :format
+
+  def title
+    format.humanize
+  end
+
+  def block_content
+    content = content_tag(:div, class: "govspeak") do
+      edition.render(embed_code_for_format)
+    end
+    content << embed_code_element
+
+    content
+  end
+
+  def embed_code_element
+    return unless edition.show_embed_codes?
+
+    content_tag(
+      :p,
+      embed_code_for_format,
+      class: "app-c-content-block-manager-format-block__embed_code",
+    )
+  end
+
+  def embed_code_for_format
+    @embed_code_for_format ||= document.embed_code_for_format(format)
+  end
+
+  def data_attributes
+    return { "testid": "format_block" } unless edition.show_embed_codes?
+
+    {
+      module: "copy-embed-code",
+      "embed-code": embed_code_for_format,
+      "embed-code-details": "format block",
+      "testid": "format_block",
+    }
+  end
+end
--- a/app/models/concerns/schema/configuration.rb
+++ b/app/models/concerns/schema/configuration.rb
@@ -2,9 +2,14 @@ class Schema
   module Configuration
     extend ActiveSupport::Concern
 
+    FORMATS_PROPERTY_KEY = "x-formats".freeze
     BLOCK_DISPLAY_FIELDS_PROPERTY_KEY = "x-block-display-fields".freeze
     EMBEDDABLE_AS_BLOCK_PROPERTY_KEY = "x-embeddable-as-block".freeze
     FIELD_ORDER_PROPERTY_KEY = "x-field-order".freeze
+
+    def formats
+      @body[FORMATS_PROPERTY_KEY] || []
+    end
 
     def block_display_fields
       @body[BLOCK_DISPLAY_FIELDS_PROPERTY_KEY] || []
--- a/app/models/schema/definitions/time_period.json
+++ b/app/models/schema/definitions/time_period.json
@@ -2,6 +2,14 @@
   "type": "object",
   "additionalProperties": false,
   "x-embeddable-as-block": true,
+  "x-formats": [
+    "long_form",
+    "months_and_years_long",
+    "start_day_and_month",
+    "start_month_as_word",
+    "years",
+    "years_short"
+  ],
   "properties": {
     "description": {
       "type": "string",
--- a/app/views/documents/show.html.erb
+++ b/app/views/documents/show.html.erb
@@ -82,6 +82,8 @@
   </div>
 <% end %>
 
+<%= render partial: "documents/formats" if @schema.formats.any? %>
+
 <% @subschemas.grouped.each do |group, subschemas| %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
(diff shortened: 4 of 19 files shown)


app/views/documents/show.html.erb
```diff
@@ -82,6 +82,19 @@
   </div>
 <% end %>
 
+<% @schema.formats.each do |format| %>
```
pezholio: Should we maybe have a subheading to demarkate where the formats start? We can probably do something a little more further down the line design wise, but this would be a start

app/models/concerns/schema/configuration.rb
```diff
@@ -2,10 +2,15 @@ class Schema
...
     extend ActiveSupport::Concern
 
+    FORMATS_PROPERTY_KEY = "x-formats".freeze
     BLOCK_DISPLAY_FIELDS_PROPERTY_KEY = "x-block-display-fields".freeze
     EMBEDDABLE_AS_BLOCK_PROPERTY_KEY = "x-embeddable-as-block".freeze
     FIELD_ORDER_PROPERTY_KEY = "x-field-order".freeze
 
+    def formats
+      @body[FORMATS_PROPERTY_KEY] || []
+    end
```
pezholio, line 12: Can we add some documentation for this new config too pls?

app/components/edition/show/format_block_component.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def embed_code_element
+    return unless edition.show_embed_codes?
+
+    content_tag(
+      :p,
+      embed_code_for_format,
+      class: "app-c-content-block-manager-format-block__embed_code",
+    )
+  end
```
graycodes, line 33: (pretty minor)
I generally expect ruby components to have their rendering responsibilities in the .html.erb file and the methods that enable this in the .rb file. I feel like it helps keep the separation of the two concerns. 
Here, if I want to see what the component is rendering I look in the .html.erb file and I can see that it's rendering a summary_card but actually there's a fair amount of rendering in the ruby too (for the contents of `block_content`).

This is probably just my preference and a completely normal thing to do in rails but I would usually expect this to maybe be in some kind of `capture` block in the html to keep the code 'cohesive' and separate the concerns. (That said, this is a component so ultimately this all relates to 'rendering' anyway...)

app/models/schema/definitions/time_period.json
```diff
@@ -2,6 +2,14 @@
...
   "additionalProperties": false,
   "x-embeddable-as-block": true,
+  "x-formats": [
+    "long_form",
+    "months_and_years_long",
+    "start_day_and_month",
+    "start_month_as_word",
+    "years",
+    "years_short"
+  ],
```
patrickjfl, line 12: are we currently maintaining two lists of formats (and maybe other metadata), one here and another in content block tools?

app/views/documents/show.html.erb
edavey: good idea, I've now added a heading before the list of available formats

app/models/concerns/schema/configuration.rb
edavey: 👍  done, thks for suggesting

app/components/edition/show/format_block_component.rb
edavey: Interesting question! I think that it's OK like this as the Ruby ViewComponent is there to support the ERB template.

 This `FormatBlockComponent` is based very closely on the existing `DefaultBlockComponent`.

They both use a SummaryCard component from PublishingComponents to render the HTML:

```erb
<%= render "govuk_publishing_components/components/summary_card", {
    title: title,
    rows: [
      {
        key: "Block content",
        value: block_content,
        data: data_attributes,
      },
    ],
  } %>
```

where as you say `block_content` is defined in FormatBlockComponent with

```rb

  def block_content
    content = content_tag(:div, class: "govspeak") do
      edition.render(embed_code_for_format)
    end
    content << embed_code_element

    content
  end

  def embed_code_element
    return unless edition.show_embed_codes?

    content_tag(
      :p,
      embed_code_for_format,
      class: "app-c-content-block-manager-format-block__embed_code",
    )
  end
```

i.e. we use [ActionView::Helpers::TagHelper#content_tag](https://apidock.com/rails/ActionView/Helpers/TagHelper/content_tag) to build both:

- the element containing the rendered format and
- the element containing the embed code for the user to copy (if published or scheduled)

The alternative as you say would be to put this markup and logic in the erb file in a capture block but that might up less readable?

app/models/schema/definitions/time_period.json
edavey: Yes this is true. If the list of available formats defined in CBM and "Tools" is out of sync:

- if CBM is missing a format: that format won't be shown in the list of "Available formats" and the embed code won't be offered

- if CBM includes a format not defined in Tools: an error will be thrown:

```
ContentBlockTools::InvalidFormatError at /42

Unknown format 'banana' for time_period
```

Having the two lists is not ideal. But once the rendering of blocks has moved from Tools to CBM as part of [ADR: Add an API for Content Block Manager](https://github.com/alphagov/content-block-manager/pull/648) then there will only be one list of formats.

edavey:
@pezholio I think this is ready for another look now. I haven't added any "feature flag" controls on the visibility of the "Available formats" as I don't see the need.

app/views/documents/show.html.erb
```diff
@@ -82,6 +82,8 @@
   </div>
 <% end %>
 
+<%= render partial: "documents/formats" %>
```
pezholio: Should we hide this partial if there are no formats? Otherwise we'll see `Available formats` on every block, even when there are no formats to view.

pezholio (changes requested):
Just one minor thing, but otherwise looks good

app/views/documents/show.html.erb
edavey: 👍 good point!

edavey: that's done now

pezholio (approved):
LGTM 👍

PR #635, opened 2026-05-06
(CM-916) Set default locale value to "en"

patrickjfl:
This fixes a bug related to a URL to view a preview of a published document, e.g. https://content-block-manager.integration.publishing.service.gov.uk/preview/a248f622-8acb-4cc0-b46e-1c39c792d243/edition/551?locale=en&state=published.

We have found that removing the “locale” value or removing the key/value pair entirely causes a 404 error. By explicitly setting the locale to “en” if the parameter value is falsy or an empty string, we avoid this error.

Jira ticket: https://gov-uk.atlassian.net/browse/CM-916.

https://github.com/user-attachments/assets/c769d739-a2a3-4ba5-b954-3ed8dc3022ea

### DIFF
--- a/engines/block_preview/app/model/block_preview/preview_content.rb
+++ b/engines/block_preview/app/model/block_preview/preview_content.rb
@@ -8,7 +8,7 @@ module BlockPreview
       @content_id = content_id
       @block = block
       @base_path = base_path
-      @locale = locale
+      @locale = locale.presence || "en"
       @state = validated_state(state)
     end
 
--- a/engines/block_preview/spec/unit/app/models/preview_content_spec.rb
+++ b/engines/block_preview/spec/unit/app/models/preview_content_spec.rb
@@ -143,6 +143,15 @@ RSpec.describe BlockPreview::PreviewContent do
       allow(BlockPreview::PreviewHtml).to receive(:new).and_return(preview_html_response)
     end
 
+    context "when the locale is empty" do
+      let(:locale) { "" }
+
+      it "defaults the locale to 'en'" do
+        preview_content.html
+        expect(BlockPreview::PreviewHtml).to have_received(:new).with(hash_including(locale: "en"))
+      end
+    end
+
     it "builds preview html using the host content base path by default" do
       expect(preview_content.html).to eq(html)
 


engines/block_preview/app/model/block_preview/preview_content.rb
```diff
@@ -5,6 +5,8 @@ class PreviewContent
     attr_reader :state
 
     def initialize(content_id:, block:, base_path: nil, locale: "en", state: "published")
+      locale = "en" if locale.blank?
+
```
pezholio: I think you can do:

```
@locale = locale || "en"
```

On line 13 instead here

edavey:
Thanks Pat, this looks good 👍 . Just to tit-pick a few little details on our ways of working: normally we:

- include the Jira ticket number in the PR title e.g. '(CM-916) Set default locale value to "en"'
- include a link to the Jira ticket in the PR narrative

Also, as our service is rather complex, we tend to invest in quite a bit of explanatory detail to assist newcomers (or ourselves) in future. So I wonder if we should flesh out the PR narrative and commit message a little to more to:

- be explicit what the "the URL" is: e.g. "a URL to view a preview of a published document"
- include an example of the URL e.g. "https://content-block-manager.integration.publishing.service.gov.uk/preview/a248f622-8acb-4cc0-b46e-1c39c792d243/edition/551?locale=en&state=published"

I appreciate that a lot of this is communicated by the video in the PR message, which is great for reviewers now, but for the future, written notes in the commit message will be king...

edavey (approved):
👍 nice work, just a few small comments on presentation and narrative.

engines/block_preview/app/model/block_preview/preview_content.rb
patrickjfl: but won't that evaluate to an empty string if `locale` is an empty string?

But yes, we could just lift this guard and put it on line 13?

patrickjfl:
> Thanks Pat, this looks good 👍 . Just to tit-pick a few little details on our ways of working: normally we:
> 
>     * include the Jira ticket number in the PR title e.g. '(CM-916) Set default locale value to "en"'
> 
>     * include a link to the Jira ticket in the PR narrative
> 
> 
> Also, as our service is rather complex, we tend to invest in quite a bit of explanatory detail to assist newcomers (or ourselves) in future. So I wonder if we should flesh out the PR narrative and commit message a little to more to:
> 
>     * be explicit what the "the URL" is: e.g. "a URL to view a preview of a published document"
> 
>     * include an example of the URL e.g. "https://content-block-manager.integration.publishing.service.gov.uk/preview/a248f622-8acb-4cc0-b46e-1c39c792d243/edition/551?locale=en&state=published"
> 
> 
> I appreciate that a lot of this is communicated by the video in the PR message, which is great for reviewers now, but for the future, written notes in the commit message will be king...

Thanks very much Ed, I've made updates as you suggest.

engines/block_preview/app/model/block_preview/preview_content.rb
pezholio: Ah, d'oh! Of course! You could use [`presence`](https://apidock.com/rails/Object/presence) in that case:

```
locale = locale.presence || "en"
```

patrickjfl: ah yes ok. will do!

PR #639, opened 2026-05-08
CM-935 create API engine

patrickjfl:
Creates the basic scaffolding and boilerplate for an API engine for us to iterate on.

Jira ticket: https://gov-uk.atlassian.net/browse/CM-935.


pezholio:
@patrickjfl I think we can bring this out of draft now the ADR is in

pezholio (approved):
LGTM 👍

PR #645, opened 2026-05-11
Add ADR to offer alternate presentations via "formats"

edavey:


docs/architecture/decisions/0012-offer-alternate-presentations-with-formats.md
```diff
@@ -0,0 +1,157 @@
+# 12. Offer alternate presentations of content blocks with formats
+
+Date: 2026-05-07
+
+Status: Draft
+
+## Context
+
+### Driver
```
pezholio, line 9: Do we want to talk about calculations here too? I feel like there's a really strong rationale here too

edavey: thanks, good point. I've added a "Calculated values" section to the examples as it seemed to fit well there...

docs/architecture/decisions/0012-offer-alternate-presentations-with-formats.md
```diff
@@ -0,0 +1,173 @@
...
+   this into the embedding document (via manual markdown) scatters
+   presentation logic into every document that uses the block.
+
+3. **Formats are declared in schemas** - Each block type's JSON schema lists
+   its supported formats via the `x-formats` extension (e.g.
+   `["income_tax_table"]` for tax, `["years_short", "start_month_as_word", ...]`
+   for time periods). This makes formats discoverable and validatable at
+   authoring time rather than failing silently at render time.
+
+4. **Fail-fast validation** - Components validate the format specifier at
```
graycodes, line 58: Ah this is cool. Does this mean the edition would fail to publish?

graycodes (approved):

graycodes:
Excellent ADR! I think you've covered a lot here with some good examples.

docs/architecture/decisions/0012-offer-alternate-presentations-with-formats.md
edavey: If I try to publish a document (e.g. via Whitehall) which includes a invalid format e.g.

```
The current tax year is {{embed:content_block_time_period:tax-year#rubbish}
```

then:

- Publishing API uses `Presenters::ContentEmbedPresenter` to call `ContentBlockTools::ContentBlock#render` for each embed code found in the document, and so
- `ContentBlockTools::ContentBlock#render` will raise `InvalidFormatError` when the invalid `rubbish` format is encountered

So yes I believe this will stop the document being published...

edavey: I've made this explicit in 

> 4. Fail-fast validation

graycodes: 🙏🏻

PR #646, opened 2026-05-11
Add draft ADR for Composite block type

edavey:


docs/architecture/decisions/0013-add-composite-block-type.md
```diff
@@ -0,0 +1,119 @@
...
+    - Avoids circular reference challenges
+
+
+## Decision
+
+Add a new block type, `composite`, which allows editors to author blocks that 
+combine free-form Govspeak text with embedded references to other content 
+blocks.
+
+### Illustrative example
```
pezholio, line 40: What's to stop people creating a "composite" block without any reference to a content block, effectively creating a "snippet" content block by the back door?

edavey: Nothing?! I guess we get the generic block for free...

pezholio: Yeah, I wonder whether we genericise the name of this and just call it a "generic" block? Might need input from IA in this case...

PR #647, opened 2026-05-11
Update adding a new schema documentation

pezholio:
There was a broken link in this Markdown file, which was removed when we unified the schema and configuration.


edavey (approved):
thanks for fixing this!

PR #648, opened 2026-05-11
ADR: Add an API for Content Block Manager

pezholio:
Document the decision to add an API to Content Block Manager to support the authoring widget's discovery/search and rendering needs.

Covers rationale (search, auth, consolidation of rendering sources), consequences (authentication strategy, replacing Content Block Tools, availability dependency, performance considerations), and suggested endpoints for search, detail, and batch render.

### DIFF
--- a/docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
+++ b/docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
@@ -1,0 +1,187 @@
+# 13. Add an API for Content Block Manager
+
+Date: 2026-05-11
+
+Status: Accepted
+
+## Context
+
+We need to make it easy for editors to use content blocks. The [**authoring widget**](https://github.com/alphagov/content-block-editor) 
+is an important piece of that strategy. The widget will be used by publishing apps to allow editors to:
+
+- preview a rendered block when faced with an embed code in their document
+- discover available blocks, preview them, and select one to have its embed code inserted into their document (e.g. 
+via an "Insert content block..." call to action)
+
+The widget needs both discovery/search and rendering to be effective. The Content Store cannot provide either 
+adequately.
+
+An API may also be useful for migrating data from other apps to the Content Block Manager, such as the planned work on 
+migrating contacts from Whitehall.
+
+## Decision
+
+We will add an API to Content Block Manager. This will use the same engine-based pattern as we already use for 
+previewing content blocks.
+
+Why we need our own API rather than relying on the Content Store:
+
+### 1. The widget needs discovery/search
+
+Content Store has no facility for searching or browsing content blocks by type, organisation, or text content.
+
+### 2. We will eventually need auth controls via Signon
+
+This is true especially for draft content blocks, which the Content Store doesn't serve at all. Authorisation will 
+restrict access based on content state and user permissions.
+
+### 3. Multiple apps currently render blocks differently, fetching from different sources
+
+Publishing API uses its own DB, Whitehall fetches from the Publishing API, Publisher and Smart Answers fetch from the 
+Content Store, and the Content Block Manager uses its local DB. A single API consolidates this fragmentation.
+
+### 4. Content Store isn't designed to serve building-block data
+
+Content store is a read-only cache of published content designed to serve JSON representations of finished pages, not 
+reusable components or rendered HTML fragments.
+
+## Consequences
+
+### Authentication
+
+Initially, we will not worry about authentication, as we intend to only surface published blocks, which are in the 
+public domain. However, if we want to surface draft blocks in the future, we may have to consider adding authentication, 
+either via an API key or OAuth authentication.
+
+If we went with API-key-based authentication, we would have to consider how we would authenticate via the client side 
+(using Content Block Editor), as we would not want to surface the API key in the frontend code.
+
+If we went with OAuth authentication, this would potentially give us greater control over what access a user has to 
+which blocks, although this would make authentication more difficult, especially when using the app cross-platform.
+
+### Replacing Content Block Tools
+
+In future, we may want to consider replacing [Content Block Tools](https://github.com/alphagov/govuk_content_block_tools) 
+as the source of truth with the API, which will allow non-Ruby applications to render content blocks.
+
+## Availability dependency
+
+If we decide to replace Content Block Tools with the API, the Publishing API (and other consumers) will depend on the 
+Content Block Manager being available. This is accepted as a trade-off for eliminating rendering duplication and 
+fragmentation.
+
+### Volume and performance
+
+Publishing apps commonly republish all their documents at once (mass republication). The API design must cope with 
+high request volumes during these events. Considerations include:
+
+- Batch vs individual embed code rendering (POST with JSON body allows batch; GET with query params is more RESTful 
+but has URL length limits for long embed codes)
+- Caching rendered output (keyed by embed code and edition version)
+- Rate limiting or queuing strategies
+
+We will propose a technical plan for caching and resilience to the GOV.UK developer and technical architecture 
+communities to get early agreement on a robust strategy.
+
+### Implementation
+
+To make documentation and interoperability easier, we should consider using OpenAPI (Swagger) to define the API. We 
+should look into tools we can use in Rails to make ensure the API is consistent and up-to-date with the specification,
+such as [Rswag](https://github.com/rswag/rswag).
+
+We should also consider using [Pact contract tests](https://docs.publishing.service.gov.uk/manual/pact-testing.html)
+within any client apps that consume the API.
+
+## Suggested endpoints
+
+### `GET /api/blocks/search`
+
+This endpoint will be used by our authoring widget to retrieve a list of blocks matching certain criteria. For example, 
+an editor wants to "Insert a block" by perusing a list of all available blocks.
+
+If no filtering criteria are supplied, the endpoint returns all available blocks. We expect that widget will have 
+a need of properties including:
+
+* organisation
+* state (draft / published)
+* block type
+* block title
+* embed code
+* available embed codes
+
+#### Parameters
+
+* `block_type` (optional)
+    * The block type to return
+* `lead_organisation` (optional)
+    * The lead organisation to filter by
+* `keyword` (optional)
+    * Any keyword(s) to filter by
+* `state` (optional)
+    * State to filter by (defaults to `published`)
+
+#### Example response
+
+```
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+### Implementation
+
+To make documentation and interoperability easier, we should consider using OpenAPI (Swagger) to define the API. We 
+should look into tools we can use in Rails to make ensure the API is consistent and up-to-date with the specification,
+such as [Rswag](https://github.com/rswag/rswag).
+
+We should also consider using [Pact contract tests](https://docs.publishing.service.gov.uk/manual/pact-testing.html)
+within any client apps that consume the API.
+
+## Suggested endpoints
```
edavey, line 95: Should we be clear on who we think will be the consumers of these endpoints? i.e. might that help us clarify the interfaces?  I'm going to suggest / try that out on each of the 3 endpoints to see if I understand correctly...

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+
+### `GET /api/blocks/search`
+
```
edavey, line 98: Suggest:

> This endpoint will be used by our authoring widget to retrieve a list of blocks matching certain criteria. For example, an editor wants to "Insert a block" by perusing a list of all available blocks.
> 
> If no filtering criteria are supplied, the endpoint returns all available blocks. We expect that widget will have need of properties including:
> 
> - organisation
> - state (draft / published)
> - block type
> - block title
> - embed code
> - available formats (as "links"?)

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+    {
+      "title": "Some content block",
+      "block_type": "pension",
+      "url": "api/content-blocks/bed722e6-db68-43e5-9079-063f623335a7",
+      "content_id": "bed722e6-db68-43e5-9079-063f623335a7"
+    }
+  ]
+}
+```
+
```
edavey, line 159: if my use-case above is correct then it suggests the response for each "result" should also include:

- state
- embed code
- available formats, perhaps as "links"?

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+      "block_type": "pension",
+      "url": "api/content-blocks/bed722e6-db68-43e5-9079-063f623335a7",
+      "content_id": "bed722e6-db68-43e5-9079-063f623335a7"
+    }
+  ]
+}
+```
+
+### `GET /api/blocks/:id`
+
```
edavey: By `:id` do we mean "content_id or embed_code"? Is there a case to be made that external users of content blocks should always reference them via their embed code only?

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+Returns information about a content block with a specific UUID. By default, returns the current public version of the
```
edavey: Suggest:

> This endpoint will be used by our authoring widget to discover full information about a particular block, for example if I hover over an embed code I receive:
> 
> - its title
> - its type
> - its organisation
> - its available state (draft, published)
> - the rendered HTML for that embed code
> - rendered HTML for the other formats of this block

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+    "title": "Some content block",
+    "block_type": "pension",
+    "details": {},
+    "embed_codes": {
+      "{{embed:content_block_pension:some-content-block}}": "HTML_RESPONSE"
+    }
+  }
+}
+```
+
```
edavey: If my use-case above is valid, then this response perhaps should also include:

- embed_code (rather than content ID if we think that makes sense)
- organisation
- state (draft or published)
- a list of other available formats for this block (e.g. `[embed_code: HTML, ...]`)

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,187 @@
...
+    "details": {},
+    "embed_codes": {
+      "{{embed:content_block_pension:some-content-block}}": "HTML_RESPONSE"
+    }
+  }
+}
+```
+
+### `POST /api/blocks/render`
+
```
edavey, line 161: Suggest:

> This endpoint will be used by publishing apps which need to replace multiple embed codes with the corresponding content blocks' HTML, e.g when publishing a document (Publishing API) or when rendering a local editor preview (Whitehall)

edavey (commented):
👍 a few suggestions on endpoints. 

I think it would be good to have a team chat about whether it makes sense to refer to blocks by their content_id or by their embed code. Embed code seems like the right approach to me but I may well be missing something important...

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
pezholio: Good point!

pezholio: The way I saw it was a way to "follow your nose" from the search result to get more information about a block. That said, I don't think there's a huge amount of value in having this, and we can just return all the info in the `/search` endpoint.

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
pezholio: I'm also thinking about the use case in the authoring widget too - we might want to add a bit of metadata about each rendered block too - maybe just the title and the block type for now

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
pezholio: I'm going to remove this endpoint for now in this case

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,175 @@
...
+Status: Accepted
+
+## Context
+
+We need to make it easy for editors to use content blocks. The [**authoring widget**](https://github.com/alphagov/content-block-editor) 
+is an important piece of that strategy. The widget will be used by publishing apps to allow editors to:
+
+- preview a rendered block when faced with an embed code in their document
+- discover available blocks, preview them, and select one to have its embed code inserted into their document (e.g. 
+via an "Insert content block..." call to action)
```
graycodes, line 14: there are some other use-cases that have recently presented themselves that I think we could add:

- Data migration
- Allowing users to author content blocks from their publishing apps
  - In this case it may be how we enable users to keep their workflow for Contacts in Whitehall whilst persisting the data to CBM.

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,175 @@
...
+## Availability dependency
+
+If we decide to replace Content Block Tools with the API, the Publishing API (and other consumers) will depend on the 
+Content Block Manager being available. This is accepted as a trade-off for eliminating rendering duplication and 
+fragmentation.
+
+### Volume and performance
+
+Publishing apps commonly republish all their documents at once (mass republication). The API design must cope with 
+high request volumes during these events. Considerations include:
```
graycodes, line 76: It's great we've noted these here and thought about these areas. I had not fully appreciated how this would position us in the critical path for document authoring (potentially). Is it worth mentioning that whilst we're considering all the options below, we're not actually implementing anything in particular yet?

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
pezholio: I think the API will be read-only for now, but I'll add something about the potential use cases for endpoints that will write too.

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,175 @@
+# 12. Add an API for Content Block Manager
```
edavey: tiny point: you'll need to make this ADR 13 as I've got in before you with the formatting ADR

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
```diff
@@ -0,0 +1,175 @@
...
+
+2. *We will eventually need auth controls via Signon* – especially for draft content blocks, which the Content Store 
+doesn't serve at all. Authorisation will restrict access based on content state and user permissions.
+
+3. **Multiple apps currently render blocks differently, fetching from different sources** – Publishing API uses its 
+own DB, Whitehall fetches from the Publishing API, Publisher and Smart Answers fetch from the Content Store, and the 
+Content Block Manager uses its local DB. A single API consolidates this fragmentation.
+
+4. *Content Store isn't designed to serve building-block data* – it's a read-only cache of published content designed 
+to serve JSON representations of finished pages, not reusable components or rendered HTML fragments.
```
edavey: Tiny point on the the formatting of the bullets: it's currently inconsistent: some are italic and some are bold. Perhaps they would be better as H3s? 

e.g

```
### 2. We will eventually need auth controls via Signon
This is especially the case for draft content blocks, which the Content Store doesn't serve at all. 
Authorisation will restrict access based on content state and user permissions.

```

docs/architecture/decisions/0013-add-an-api-for-content-block-manager.md
edavey: Yes, good point, I think we ought to note something brief along the lines of

> We will propose a technical plan for caching and resilience to the GOV.UK developer and technical architecture communities to get early agreement on a robust strategy.

docs/architecture/decisions/0012-add-an-api-for-content-block-manager.md
pezholio: Sorted 👍

edavey (approved):

PR #653, opened 2026-05-13
Update lint take task to autocorrect `.erb` files in all environments

patrickjfl:
We want to be able to autocorrect files locally rather than just in dev.

### DIFF
--- a/lib/tasks/lint.rake
+++ b/lib/tasks/lint.rake
@@ -2,11 +2,8 @@
 desc "Run all linters"
 task :lint do
   sh "bundle exec rubocop"
-  if Rails.env.development?
-    sh "bundle exec erb_lint --lint-all --autocorrect"
-  else
-    sh "bundle exec erb_lint --lint-all"
-  end
+  sh "bundle exec erb_lint --lint-all"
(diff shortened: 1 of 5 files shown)


lib/tasks/lint.rake
```diff
@@ -2,11 +2,7 @@
 desc "Run all linters"
 task :lint do
   sh "bundle exec rubocop"
```
pezholio, line 4: Do we want to autocorrect Rubocop too? Maybe there's a case for a `rake lint:autocorrect` command that autocorrects all linters and a `rake lint` command that just runs the linters?

pezholio:
Would also be good for this to be documented in the `README`. For some reason I didn't know this command was here!

lib/tasks/lint.rake
patrickjfl: oh yeah, good point. will update on all fronts!

graycodes: there's also a `lint:prettier:fix` in the package.json we can add I reckon

pezholio (approved):

graycodes (approved):
mega

PR #672, opened 2026-05-19
(CM-951) Search endpoint for API

pezholio:
This adds a new search endpoint to allow users to search for content blocks by a number of criteria. I've used [Rswag](https://github.com/rswag/rswag) to generate documentation based on our Rspec tests. There's quite a few commits here, but I've tried to build up the functionality bit by bit, so most of the commits should be fairly small.

### DIFF
--- a/.github/workflows/check-open-api.yml
+++ b/.github/workflows/check-open-api.yml
@@ -1,0 +1,41 @@
+name: Check OpenAPI documentation is up to date
+on:
+  workflow_call:
+
+jobs:
+    check-open-api:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+          with:
+            show-progress: false
+
+        - name: Setup Ruby
+          uses: ruby/setup-ruby@6aaa311d81eba98ae12eaffbcb63296ace0efcde # v1.307.0
+          with:
+            bundler-cache: true
+
+        - name: Setup Postgres
+          id: setup-postgres
+          uses: alphagov/govuk-infrastructure/.github/actions/setup-postgres@main
+
+        - name: Initialize database
+          env:
+            RAILS_ENV: test
+            TEST_DATABASE_URL: ${{ steps.setup-postgres.outputs.db-url }}
+          run: bundle exec rails db:setup
+
+        - name: Build OpenAPI documentation
+          run: bundle exec rake api:generate_swagger
+          env:
+            RAILS_ENV: test
+            TEST_DATABASE_URL: ${{ steps.setup-postgres.outputs.db-url }}
+
+        - name: Check for uncommitted changes
+          run: |
+            if git diff --exit-code; then
+              echo "No uncommitted changes detected."
+            else
+              echo "::error title=Uncommitted changes to OpenAPI docs::If these are your changes, run \\\`rake api:generate_swagger\\\` and commit the changes."
+              exit 1
+            fi
--- a/app/public/queries/content_block/query.rb
+++ b/app/public/queries/content_block/query.rb
@@ -1,9 +1,19 @@
 class ContentBlock
   class Query
-    def self.call
+    module Scopes
+      def by_block_type(block_type)
+        return self if block_type.blank?
+
+        where("block_type = ?", block_type)
+      end
+    end
+
+    def self.call(filters = {})
       Document.joins(:editions)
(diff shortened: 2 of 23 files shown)


engines/api/features/api/step_definitions/api_steps.rb
```diff
@@ -0,0 +1,7 @@
+When("I access the API endpoint {string}") do |url|
+  visit url
+end
+
+Then("the response should have status code {int}") do |status_code|
+  expect(page.status_code).to eq(status_code)
+end
```
patrickjfl: nice!

spec/swagger_helper.rb
```diff
@@ -0,0 +1,32 @@
+require "spec_helper"
+
+RSpec.configure do |config|
+  config.openapi_root = Rails.root.join("engines/api/swagger").to_s
+
+  config.openapi_specs = {
```
patrickjfl, line 6: Could you add some notes in the commit message about what this is and how it works?

engines/api/spec/requests/api_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+    get "Search content blocks" do
+      tags "Content Blocks"
+      produces "application/json"
+
+      response "200", "blocks found" do
+        run_test!
+      end
+    end
+  end
+end
```
patrickjfl, line 222: So is the swagger documentation built up from these tests?

spec/unit/app/public/queries/content_block/query_spec.rb
```diff
@@ -0,0 +1,23 @@
+RSpec.describe ContentBlock::Query do
+  describe ".call" do
+    it "returns one content block per document" do
+      doc_a = create(:document)
+      create(:edition, :published, document: doc_a, updated_at: 2.days.ago)
```
patrickjfl: bit nit picky but could you give this one the title "Doc A old edition" for clarity?

app/public/queries/content_block/query.rb
```diff
@@ -6,6 +6,12 @@ def by_block_type(block_type)
 
         where("block_type = ?", block_type)
       end
+
+      def by_lead_organisation_id(lead_organisation_id)
+        return self if lead_organisation_id.blank?
```
patrickjfl, line 53: what does this line mean? Are we including the block in the results if it has no lead org ID?

patrickjfl: If so then I think we should probably have a test that documents that behaviour?

patrickjfl (commented):
I'm gonna have to come back to this because it's too much for my head to take in in one go. Looks like a lot of really solid work though, well done Stu! I've added a few comments on things I wasn't clear on.

engines/api/spec/requests/api_spec.rb
pezholio: Yep 👍

app/public/queries/content_block/query.rb
pezholio: The `Scopes` module extends the `Document` query, which allows us to chain queries together. If `lead_organisation_id` is blank, then we just return the query, otherwise, we chain our `where("block_type = ?", block_type)` to the query and return that.

spec/unit/app/public/queries/content_block/query_spec.rb
pezholio: ✅

app/public/queries/content_block/query.rb
pezholio: The Thoughtbot article mentioned in the previous commit probably covers this more eloquently than I could https://thoughtbot.com/blog/a-case-for-query-objects-in-rails#a-better-query-object

app/public/queries/content_block/query.rb
```diff
@@ -1,6 +1,7 @@
 class ContentBlock
   class Query
     DEFAULT_PAGE_SIZE = 10
+    Result = Data.define(:blocks, :current_page, :total_pages, :total_count)
```
patrickjfl, line 4: cool 😎

.github/workflows/check-open-api.yml
```diff
@@ -0,0 +1,41 @@
+name: Check OpenAPI documentation is up to date
+on:
+  workflow_call:
```
patrickjfl, line 3: nice

patrickjfl (approved):
Looks good to me, well done for all this Stu 🚀

PR #674, opened 2026-05-20
API engine missing from docker build

graycodes:
This just ensures that the directory is available in docker so that it can be used.

Currently failing the build : https://github.com/alphagov/content-block-manager/actions/runs/26160252231/job/76949999382

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,7 @@ ENV JWT_AUTH_SECRET=unused_yet_required
 
 WORKDIR $APP_HOME
 COPY Gemfile* .ruby-version ./
-COPY engines/fact_check ./engines/fact_check
-COPY engines/block_preview ./engines/block_preview
+COPY engines ./engines
 RUN bundle install
 COPY package.json yarn.lock ./
 RUN yarn install --production --frozen-lockfile --non-interactive --link-duplicates


Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR $APP_HOME
 COPY Gemfile* .ruby-version ./
 COPY engines/fact_check ./engines/fact_check
 COPY engines/block_preview ./engines/block_preview
+COPY engines/api ./engines/api
```
pezholio: I wonder if there's a way to copy every directory from `engines`?

pezholio (approved):
LGTM 👍

Dockerfile
graycodes: yeah this was one of those things that went through my mind but didn't land... I've fixed it now to just copy the whole engines dir!

Dockerfile
```diff
@@ -8,8 +8,7 @@ ENV JWT_AUTH_SECRET=unused_yet_required
 
 WORKDIR $APP_HOME
 COPY Gemfile* .ruby-version ./
-COPY engines/fact_check ./engines/fact_check
-COPY engines/block_preview ./engines/block_preview
+COPY engines ./engines
```
pezholio, line 11: Nice 👍

pezholio (approved):

PR #679, opened 2026-05-21
Various tweaks to our new API

edavey:
Please see individual commits for details. The most visible change is the inclusion of the RSwag UI library to serve up our OpenAPI spec at `/api-docs`:

open_api_spec_served


engines/api/app/presenters/api/results_presenter.rb
```diff
@@ -55,7 +55,7 @@ def self_link
     end
 
     def page_href(offset)
-      request_url.tap { |url|
+      request_url.dup.tap { |url|
```
patrickjfl, line 58: very nice and good spot!

app/public/queries/content_block/query.rb
```diff
@@ -63,7 +63,7 @@ def by_lead_organisation_id(lead_organisation_id)
       def by_keyword(keyword)
         return self if keyword.blank?
 
-        where("documents.id IN (?)", Document.with_keyword(keyword).pluck(:id))
+        where(id: Document.with_keyword(keyword).select(:id))
```
patrickjfl, line 66: 🎉

engines/api/spec/requests/api_spec.rb
```diff
@@ -226,6 +226,41 @@
...
+
+        response "400", "returns error for zero page number", document: false do
+          let(:page) { 0 }
+
+          run_test! do |response|
+            data = JSON.parse(response.body)
+            expect(data["error"]).to be_present
+          end
+        end
+      end
```
patrickjfl, line 263: should we be checking the status code here? or is that Rails magic that comes from the `response "400"` declaration at the start of the test definition?

engines/api/features/api/step_definitions/api_steps.rb
```diff
@@ -8,9 +8,10 @@
 end
 
 Given(/^there are the following published content blocks:$/) do |table|
-  table.hashes.each do |hash|
+  table.hashes.each_with_index do |hash, index|
     hash["lead_organisation_id"] = Organisation.all.find { |org| org.name == hash["organisation"] }.id
     hash["document"] = create(:document, block_type: hash["block_type"])
+    hash["created_at"] = index.days.ago
```
patrickjfl, line 14: hmmmm, this seems like a similar case to another flaky test I fixed earlier: https://github.com/alphagov/content-block-manager/pull/688/changes/4ff97786c6111f65238b12342997f5b8c7da8a9b. I assumed that the order of the results didn't matter in that case, but perhaps it does?

patrickjfl (approved):
Brilliant work Ed, well done.

engines/api/spec/requests/api_spec.rb
edavey: yes the `response` method asserts the HTTP response status. See

https://github.com/rswag/rswag/blob/master/rswag-specs/lib/rswag/specs/example_group_helpers.rb#L66-L68

If there's a mismatch between the expected response and the actual response we'll see an error like:
```
Rswag::Specs::UnexpectedResponse:
       Expected response code '400' to match '401'
```

engines/api/features/api/step_definitions/api_steps.rb
edavey: It looks like you've correctly fixed up that spec to do what it says:

`it "returns one content block per document" do`

We've another spec which describes the ordering:

https://github.com/alphagov/content-block-manager/blob/5dec2e12230b28cf83593cd5b7399dc740c905c3/spec/unit/app/public/queries/content_block/query_spec.rb#L77

PR #689, opened 2026-05-28
CM-944: store pension rate amount as numeric string

patrickjfl:
Jira [CM-944](https://gov-uk.atlassian.net/browse/CM-944)

This PR makes breaking changes to the pension JSON schema, removing the expectation that the pension "amount" field will contain a "£" prefix and instead adding a "x-input-prefix" field to the pension schema amount field to pass through a configurable prefix separate from the value. This will allow us to make better and easier use of the pension value as a number (in e.g. pension calculations), as well as making use of the existing [input with a prefix GDS component](https://design-system.service.gov.uk/components/text-input/input-prefix-suffix/) in our pension block form (included within this PR).

### **As these are breaking changes, they should not be merged until we are satisfied that we have a plan for migrating existing pension amount values via a Rake task.**

Screenshot of changes:
Screenshot 2026-05-28 at 11 52 27

### Tools gem updates
Confirm that the following users of content_block_tools are running v.1.14 so they have the new [PensionRatePresenter](https://github.com/alphagov/govuk_content_block_tools/pull/160) which copes with both forms of rate:

"£100.30" and
"100.30":

- [x] Whitehall
- [x] Publisher
- [x] Publishing-API
- [x] Smart answers

## End to end test suites

Verify that both E2E suites will pass (or plan a sequence of expected failures and fixes):

- [x] GOV.UK E2E ([alphagov/govuk-e2e-tests](https://github.com/alphagov/govuk-e2e-tests/blob/main/tests/content-block-manager.spec.js))
- [x] Content Modelling E2E ([alphagov/content-modelling-e2e](https://github.com/alphagov/content-modelling-e2e/blob/main/tests/content-block-manager.spec.js))

[CM-944]: https://gov-uk.atlassian.net/browse/CM-944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/edition/details/fields/string_component.rb
+++ b/app/components/edition/details/fields/string_component.rb
@@ -7,5 +7,5 @@ private
 
   attr_reader :context
 
-  delegate :label, :name, :id, :value, :error_items, :hint_text, to: :context
+  delegate :label, :name, :id, :value, :error_items, :hint_text, :input_prefix, to: :context
 end
--- a/docs/schemas.md
+++ b/docs/schemas.md
@@ -53,19 +53,19 @@ Using the pension schema as an example, the `rates` payload looks like this:
   "rates": {
     "rate-1": {
       "title": "Rate 1",
-      "amount": "£221.20",
+      "amount": "221.20",
       "frequency": "a week",
       "description": "Your weekly pension amount"
     },
     "rate-2": {
       "title": "Rate without decimal point",
-      "amount": "£221",
+      "amount": "221",
       "frequency": "a week",
       "description": "Your weekly pension amount"
     },
     "rate-3": {
       "title": "Rate with big value",
-      "amount": "£1,223",
+      "amount": "1,223",
       "frequency": "a week",
       "description": "Your weekly pension amount"
     }
--- a/engines/fact_check/spec/components/ungrouped_subschema_diff_component_spec.rb
+++ b/engines/fact_check/spec/components/ungrouped_subschema_diff_component_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe FactCheck::UngroupedSubschemaDiffComponent, type: :component do
                            {
                              "title" => "",
                              "frequency" => "",
-                             "amount" => "",
+                             "amount" => { "type" => "string", "x-input-prefix" => "£" },
                            } })
   end
   let(:schema) { build(:schema) }
@@ -18,7 +18,7 @@ RSpec.describe FactCheck::UngroupedSubschemaDiffComponent, type: :component do
               "my_rate" => {
                 "title" => "my title",
                 "frequency" => "weekly",
-                "amount" => "£1000",
+                "amount" => "1000",
               },
               "my_rate_2" => {
                 "title" => "my title 2",
--- a/lib/tasks/remove_pound_symbol_prefix_from_pension_amount_value.rake
+++ b/lib/tasks/remove_pound_symbol_prefix_from_pension_amount_value.rake
@@ -1,0 +1,44 @@
+namespace :data_migration do
+  desc "Remove pound symbol prefix from Pension#amount values"
+  task remove_pound_symbol_prefix_from_pension_amount_value: :environment do
+    log = proc { |str|
+      timestamp = Time.current.strftime("%F %T")
+      line = "#{timestamp} > #{str}"
+      puts line unless Rails.env.test?
+      Rails.logger.info(line)
+    }
+
+    log.call("Removing pound symbol prefix from Pension#amount values")
+    editions = Edition.joins(:document).where(document: { block_type: "pension" })
+
+    log.call("#{editions.count} editions of type Pension found")
+
+    editions.each do |edition|
+      rates = edition.details["rates"]
+      updated_any_rate = false
+
+      next unless rates
+
+      rates.each do |_, rate|
+        amount = rate["amount"]
+
+        next unless amount&.start_with?("£")
+
+        log.call("Converting edition #{edition.id}: #{edition.title} (#{rate['amount']} #{rate['frequency']})")
+
+        rate["amount"] = amount.delete_prefix("£")
+        updated_any_rate = true
+      end
+
+      next unless updated_any_rate
+
+      edition.details["rates"] = rates
+      edition.save!(validate: false)
+      log.call(" -> Edition #{edition.id} successfully updated.")
+
+      edition.details["rates"].each do |_, rate|
+        log.call("Edition #{edition.id} final rate value: #{rate['amount']} #{rate['frequency']}")
+      end
+    end
+  end
+end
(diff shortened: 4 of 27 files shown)


edavey:
we need to ensure that all applications using the "tools" gem are upgraded to v.1.14 so they have the new [PensionRatePresenter ](https://github.com/alphagov/govuk_content_block_tools/pull/160)which copes with both forms of rate:

- `"£100.30"` and
- `"100.30"`

patrickjfl:
> we need to ensure that all applications using the "tools" gem are upgraded to v.1.14 so they have the new [PensionRatePresenter ](https://github.com/alphagov/govuk_content_block_tools/pull/160)which copes with both forms of rate:
> 
>     * `"£100.30"` and
> 
>     * `"100.30"`

good point, I'll look into this, thanks.

spec/unit/app/models/schema/field/configuration_spec.rb
```diff
@@ -153,4 +153,22 @@
...
+      it "returns the input prefix" do
+        expect(field.input_prefix).to eq("£")
+      end
+    end
+
+    context "when the body does not include an `x-input-prefix` property" do
+      it "returns nil" do
+        expect(field.input_prefix).to be_nil
+      end
+    end
```
edavey, line 175: I wonder if we should make this clearer? At present it relies on a knowledge of the setup of `field` all the way up at line 3.

If we were use a similar context block as you use in the first context, then the difference in the two scenarios might be clear at a glance, e.g.

```rb
    context "when the body does not include an `x-input-prefix` property" do
      let(:body) do
        { "properties" => { "something" => { "type" => "string" } } }
      end

      it "returns nil" do
        expect(field.input_prefix).to be_nil
      end
    end
```

app/components/edition/details/fields/string_component.rb
```diff
@@ -7,5 +7,5 @@ def initialize(context)
 
   attr_reader :context
 
-  delegate :label, :name, :id, :value, :error_items, :hint_text, to: :context
+  delegate :label, :name, :id, :value, :error_items, :hint_text, :input_prefix, to: :context
```
edavey, line 10: should we add a context to  `spec/components/edition/details/fields/string_component_spec.rb` to describe how this method is available?

something like:

```rb
  context "when an input prefix is defined" do
    before do
      expect(context).to receive(:input_prefix).and_return("£")
    end

    it "should render the input prefix" do
      render_inline component

      expect(page).to have_content "£"
    end
  end
```

spec/components/shared/embedded_objects/summary_card_component_spec.rb
```diff
@@ -168,6 +168,72 @@
...
+    component = described_class.new(
+      edition:,
+      object_type: "embedded-objects",
+      object_title: "my-embedded-object",
+    )
+
+    render_inline component
+
+    expect(page).to have_css ".govuk-summary-list__value", text: "100.00"
+  end
```
edavey, line 243: nice work, describing the 3 permutations. Did you think about using contexts for each of those? That might make the spec easier to scan? And nesting all 3 contexts in a describe block? e.g.

```rb
  describe "input prefix (e.g. for currency symbol)" do
    context "when the field has an input prefix" do
      it "renders a value with a prefix" do 
    end
    
    context "when the field has an empty input prefix" do
      it "renders a value with no prefix" do 
    end
    
    context "when the field has no input prefix" do
      it "renders a value with no prefix" 
    end
  end
```

Whatever we can do to make it as clear and descriptive as possible!

docs/schemas.md
```diff
@@ -53,19 +53,19 @@ Using the pension schema as an example, the `rates` payload looks like this:
...
-      "amount": "£221",
+      "amount": "221",
       "frequency": "a week",
       "description": "Your weekly pension amount"
     },
     "rate-3": {
       "title": "Rate with big value",
-      "amount": "£1,223",
+      "amount": "1,223",
       "frequency": "a week",
```
edavey, line 69: as this is documentation rather than a test perhaps you should either:

- break this out into a separate commit, or
- change the commit message to refer to "tests and documentation"

engines/fact_check/spec/components/ungrouped_subschema_diff_component_spec.rb
```diff
@@ -4,7 +4,7 @@
                            {
                              "title" => "",
                              "frequency" => "",
-                             "amount" => "",
+                             "amount" => { "type" => "string", "x-input-prefix" => "£" },
```
edavey, line 7: hmm, does this addition of the £ belong here?

lib/tasks/remove_pound_symbol_prefix_from_pension_amount_value.rake
```diff
@@ -0,0 +1,46 @@
+# :nocov:
```
edavey: I think we do want to unit test this, and so should remove the `:nocov:` directive...

edavey (commented):
Very clear 👍 

I've made a few minor suggestions.

And let's add unit tests to the data migration just in case that helps us see anything else...

edavey:
There's also

> 6. Update Playwright E2E tests

to consider, before we merge this.

edavey:
> There's also
> 
> > 6. Update Playwright E2E tests
> 
> to consider, before we merge this.

I've added another pre-merge checklist to the main PR narrative!

engines/fact_check/spec/components/ungrouped_subschema_diff_component_spec.rb
patrickjfl: to be honest, I have no certainty that it does. But the test at the bottom fails without it:

`expect(blocks_components[0]).to have_content("£1000")`

if we're testing the rendering of a pension component in this file then surely we should include the prefix?

spec/unit/app/models/schema/field/configuration_spec.rb
patrickjfl: good point, have updated

app/components/edition/details/fields/string_component.rb
patrickjfl: thanks, have done

spec/components/shared/embedded_objects/summary_card_component_spec.rb
patrickjfl: yes, much better!

engines/fact_check/spec/components/ungrouped_subschema_diff_component_spec.rb
edavey: and the test passes if you roll back to the previous commit?

patrickjfl: yes, maybe because we're building the edition with the "£" in the string and it's not being validated?

edavey (approved):

PR #690, opened 2026-05-28
Fix flaky query spec test

patrickjfl:
The "returns one content block per document" test was flaky due to both published editions using Time.current as their created_at timestamp. I have fixed this but also generally refactored the tests so that we have more tests, with hopefully more descriptive descriptions, and fewer assertions within each test. This should make it easier to debug when a test fails.

### DIFF
--- a/spec/unit/app/public/queries/content_block/query_spec.rb
+++ b/spec/unit/app/public/queries/content_block/query_spec.rb
@@ -1,27 +1,26 @@
 RSpec.describe ContentBlock::Query do
   describe ".call" do
-    it "returns one content block per document" do
-      doc_a = create(:document)
-      create(:edition, :published, document: doc_a, updated_at: 2.days.ago, title: "Doc A old edition")
-      create(:edition, :published, document: doc_a, updated_at: 1.day.ago, title: "Doc A new edition")
-
-      doc_b = create(:document)
-      create(:edition, :published, document: doc_b, updated_at: Time.current, title: "Doc B new edition")
-
-      excluded_document = create(:document)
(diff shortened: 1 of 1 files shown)


spec/unit/app/public/queries/content_block/query_spec.rb
```diff
@@ -20,8 +20,29 @@
...
+      result = described_class.call
+
+      expect(result.current_page).to eq(1)
+      expect(result.total_pages).to eq(1)
+      expect(result.total_count).to eq(3)
+
+      expect(result.blocks).to all(be_a(ContentBlock))
+
+      expect(result.blocks.size).to eq(3)
+      expect(result.blocks.map(&:title)).to eq(["Doc B edition", "Doc C edition", "Doc A edition"])
```
edavey: where you are making a number of assertions within a single "example" you might consider wrapping the assertions in an `aggregate_failures do...` block, see https://rspec.toolboxforweb.xyz/docs/rspec-core/expectation_framework_integration/aggregating_failures 

Or you might consider whether this is actually nudging you toward a more more granular set of examples. e.g. 

```
describe ".call"
  it "includes the number of the current page"
  it "includes the count of total pages"
  it "includes a count of the total blocks found"
  it "excludes draft content blocks"
  it "returns blocks as ContentBlock representations"
  it "orders blocks newest first"
```

edavey (approved):
Nice work breaking this out into a more granular specification. I think you _could_ go further down this road...?

edavey:
Nice work, this should be easier to maintain and change now...

edavey (approved):

PR #692, opened 2026-06-02
Add an API endpoint to render embed codes

tobybatch:


engines/api/spec/requests/api_spec.rb
```diff
@@ -274,4 +274,71 @@ def it_returns_correct_pagination_information(data, total:, pages:, current_page
       expect(data["links"][index]["href"]).to eq(link[:href])
     end
   end
+
+  let(:organisations) { build_list(:organisation, 1) }
+
+  describe "GET /api/blocks/render" do
```
edavey: This gives a clear description of the API behaviour, but the style of testing we've started on is using [RSwag](https://github.com/rswag/rswag) (see Stu's description of the `/blocks/search` endpoint at the top of this file).

The advantage of this is that when the test is run it generates our OpenAPI / Swagger spec at `engines/api/swagger/v1/swagger.json` (also available to view at `api-docs/index.html`). Ensuring that the "example"  returned in the OpenAPI spec is realistic also helps to drive out maximum clarity:

Image

app/public/models/content_block.rb
```diff
@@ -10,6 +10,13 @@ def from_edition_id(edition_id)
       edition = Edition.find(edition_id)
       ContentBlock.new(edition)
     end
+
+    def from_embed_code(embed_code)
+      document = Document.find_by(embed_code:)
+      return nil unless document&.most_recent_edition
+
+      ContentBlock.new(document.most_recent_edition)
```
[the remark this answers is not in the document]
edavey: hmm. this is a good point which we haven't thought through yet. The default behaviour should be to show the `Document#latest_published_edition`.  

A bit further down the road we might well need a mechanism to indicate if there is also a draft version of the block available. Though not everyone would necessarily have access to that draft version.

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -6,6 +6,27 @@ def search
     render json: Api::ResultsPresenter.present(result, request.original_url)
   end
 
+  # named _render to avoid collision with the core framework render method
+  def _render
+    embed_codes = params.permit(embed_codes: [])[:embed_codes] || []
+
```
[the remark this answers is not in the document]
edavey: Yes I agree with this. We don't use leading underscores much in Ruby/Rails (except to signpost unused method arguments and block variables).

engines/api/features/api/render.feature
```diff
@@ -0,0 +1,29 @@
+Feature: /api/blocks/render
+  Background:
+    Given the organisation "Ministry of Example" exists
+    And there are the following published content blocks:
+      | title               | organisation        | block_type |
+      | State Pension block | Ministry of Example | pension    |
+      | Tax year block      | Ministry of Example | pension    |
+
```
edavey, line 8: - the pension block will have the block type "pension", and 
- the tax year block will be of type "time_period"

If we're going to describe the behaviour when the given embed code is qualified by an "internal content path" then we should probably set up a "contact" block also? See below...

engines/api/features/api/render.feature
```diff
@@ -0,0 +1,29 @@
...
+      | embed_code                |
+      | {{embed:unknown}}         |
+    Then the response contains 0 rendered blocks
+
+  Scenario: Render works with embed code variants
+    When I query the render API endpoint for the following content blocks with variants:
+      | title               | variant    |
+      | State Pension block | #section-a |
+    Then the response contains 1 rendered blocks
+    And the rendered block for "State Pension block" includes title "State Pension block"
```
edavey, line 29: By "variant" I think you mean "format"? A valid format for a time period is `years_short` (see https://github.com/alphagov/content-block-manager/blob/412e6ddef98a2d25c33354934420918fb0bc5abe/app/models/schema/definitions/time_period.json#L5-L12)

We should assert that we see the expected format, e.g. ("2026-27")

Also: I think we need a scenario to describe the behaviour when the given embed code includes an "internal content path" segment e.g.

```
{{embed:content_block_contact:pension-office/contact_links/new-state-pension}}
```

as distinct from the full contact block:

```
{{embed:content_block_contact:pension-office}}
```

Again we should probably assert that our HTML snippet includes on the expected "contact link".

Your contact fixture for the above might have `#details`  something like:

```
 {
 	"description" => "Contact info for new state pension",
 	"addresses" => {
 		"address" => {"title" => "Address", "street_address" => "6 the Street"}
 	},
 	"contact_links" => {
 		"new-state-pension" => {
 		"title" => "New state pension", "label" => "webpage", "url" => "https://example.com/new_state_pension"
 		}
 	}
 }
```

edavey (commented):
Nice work, thanks for getting up and running so fast. This is a fairly far-ranging feature, so I think it's inevitable that there are a few aspects which you couldn't appreciate on week 2!

A few initial comments here before I knock off for the evening.

thks! 🙇

PR #703, opened 2026-06-04
Added an API endpoint to return rendered blocks

tobybatch:
Adds and endpoint mounted at `/api/blocks/render` that takes a list of embed codes and returns the rendered html for them. It is intended to work with [CM-966](https://github.com/alphagov/content-block-editor/pull/119)

This PR replaces https://github.com/alphagov/content-block-manager/pull/692

[CM-966]: https://gov-uk.atlassian.net/browse/CM-966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/public/models/content_block.rb
```diff
@@ -10,6 +10,18 @@ def from_edition_id(edition_id)
...
+    # ContentBlock.from_embed_code uses most_recent_edition, which includes drafts. Because the API engine controllers
+    # are unauthenticated (ActionController::API), this would allow the new /api/blocks/render endpoint to render draft
+    # content blocks if someone knows an embed code. Use the latest published edition instead so the endpoint only
+    # exposes published content.
+    # https://github.com/alphagov/content-block-manager/pull/703#discussion_r3361821179
+    def from_embed_code(embed_code)
+      document = Document.find_by(embed_code:)
+      return nil unless document&.most_recent_edition
+
+      ContentBlock.new(document.most_recent_edition)
```
edavey: As I think we noted on the other PR, initially this needs to be the `Document#latest_published_edition` rather than the latest edition. In due course we'll probably want to give editors the ability to toggle between the published block and a draft (if one exists) -- but that will require an authorisation test, as not everyone should have access to all draft blocks.

edavey (commented):
Thks!

An immediate comment regarding "latest edition" versus "latest published edition".

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -1,11 +1,36 @@
 class Api::BlocksController < Api::ApplicationController
-  def search
+  def blocks_search
```
edavey, line 2: I think certainly belongs in it's own preliminary commit? 

I'm guessing that the narrative you might capture here is along the lines of:

- we can't have the method named `render`
- let's use `blocks_render`
- as a preliminary we'll change `search` to have a similar form

By breaking out the work into multiple atomic commits we are able to provide a narrative to communicate clearly why each change is made. i.e. the responsibility is for the author to present clearly rather than for the reader to figure out.

Please have a look at https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/working-with-git.html#atomic-commits and https://www.annashipman.co.uk/jfdi/good-pull-requests.html#make-the-pull-request-tell-a-story

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -1,11 +1,36 @@
...
     render json: Api::ResultsPresenter.present(result, request.original_url)
   end
 
+  def blocks_render
+    parsed_query = Rack::Utils.parse_query(request.query_string)
+    embed_codes = (
+      Array(parsed_query["embed_codes[]"]) +
+      Array(parsed_query["embed_codes"]) +
+      Array(params[:embed_codes])
+    ).filter_map(&:presence).uniq
```
edavey, line 15: It looks like we're handling 3 different  ways of passing a query string

However as our test at `engines/api/spec/requests/api_spec.rb` is only exercising one form:

`embed_codes[]`

it's not clear to me that all this query string processing is necessary. If it is then probably we should document why in our tests, e.g.

```
context "when embed code passed as ..."
  # set up query string type a
  response "200", "the matching block is rendered"
end

context "when embed code passed as ..."
  # set up query string type a
  response "200", "the matching block is rendered"
end

context "when embed code passed twice.."
  # set up duplicated embed code in query string
  response "200", "the matching block is rendered once only"
end
```

engines/api/features/api/step_definitions/api_steps.rb
```diff
@@ -60,3 +67,85 @@
...
+Then("the response contains {int} rendered blocks") do |count|
+  expect(rendered_blocks.count).to eq(count)
+end
+
+Then("the rendered block for {string} includes title {string}") do |lookup, expected_title|
+  rendered_block = rendered_block_for(lookup)
+
+  expect(rendered_block).to include("title" => expected_title)
+  expect(rendered_block["html"]).to be_a(String)
+  expect(rendered_block["html"]).to include(expected_title)
```
edavey, line 110: I don't think these 2 final assertions are adding anything are they? Line 108 already asserts the presence of the expected title?

engines/api/features/api/step_definitions/api_steps.rb
```diff
@@ -60,3 +67,85 @@
...
+
+def rendered_blocks
+  @render_body.fetch("rendered_blocks")
+end
+
+def rendered_block_for(lookup)
+  return rendered_blocks[lookup] if rendered_blocks.key?(lookup)
+
+  title_match = rendered_blocks.values.find { |block| block["title"] == lookup }
+  return title_match if title_match
```
edavey, line 131: I'm a bit confused by this method as the name `rendered_block_for(lookup)` doesn't prepare me for the fact that it has special behaviour when the look up matches on the `title`. In other cases it delegates to `find_by_partial_key(lookup)` which I find _really_ complex!

engines/api/features/api/render.feature
```diff
@@ -0,0 +1,61 @@
...
+      | feature    |
+      | #section-a |
+    Then the response contains 1 rendered blocks
+    And the rendered block for "State Pension block" includes title "State Pension block"
+
+  Scenario: Render html for Time period content blocks
+    Given there are the following published content blocks:
+      | title            | organisation        | block_type  | details                                                                          |
+      | Test time period | Ministry of Example | time_period | {"date_range": {"start": "2026-11-05T00:00:00Z", "end": "2028-02-22T23:59:59Z"}} |
+    When I query the render API endpoint for "Test time period" with the following features or paths:
```
edavey, line 35: by "feature" I think you mean "format"?

engines/api/features/api/step_definitions/api_steps.rb
```diff
@@ -60,3 +67,85 @@
...
+
+When("I query the render API endpoint for the following content blocks:") do |table|
+  visit_render_api(table.hashes.map { |hash| published_content_blocks_by_title.fetch(hash["title"]).embed_code })
+end
+
+When("I query the render API endpoint for {string} with the following features or paths:") do |title, table|
+  embed_code = published_content_blocks_by_title.fetch(title).embed_code
+  embed_codes = table.hashes.map do |hash|
+    feature = hash["feature"]
+    feature.present? ? embed_code.sub(/(\}\}+\z)/, "#{feature}\\1") : embed_code
```
edavey, line 83: I _think_ that "feature" here stands for "format"? But I don't quite follow what's going on with `embed_code.sub(/(\}\}+\z)/, "#{feature}\\1")`

Gemfile.lock
edavey, line 1056: this doesn't belong here (I'm not sure why this keeps popping up!)

edavey (commented):
I've made a number of comments inline and I summarise my thoughts here. Finally I suggest we actually implement something simpler here. Apologies for sending you round the houses!


## GET rather than POST

[CM-952](https://gov-uk.atlassian.net/browse/CM-952) describes a POST with a list of embed codes in the body. You've implemented a GET with a list of embed codes in a query string. It's fine to deviate from the plan of course but it would be helpful to be read about your decision and rationale in the PR narrative or in the Jira ticket.

## PR structure

This PR would benefit from being broken down into a number of smaller commits. That would make it easier to review, meaning: a) quicker review and b) better review. Apologies for not emphasising this more clearly earlier, but this really is a requirement on this project. The idea, in a nutshell, is to walk the reader through the process in easy steps. Please see https://www.annashipman.co.uk/jfdi/good-pull-requests.html#make-the-pull-request-tell-a-story

I could imagine a sequence of commits something like:

- preliminary refactoring of `search` to `blocks_search`
- controller and unit tests
    - single code in url param
    - one or many embed codes in form body
- feature spec to describe behaviour when
    - supplying a format
    - supplying an internal content path

## Feature test

There's some really gnarly "helper" code in `api_steps.rb`. As our cucumber features are intended to be helpful descriptions of how the service behaves we need to keep them simple -- easy to understand and adapt as the behaviour of the service evolves.

## Rethink and simplification

Rather than take a 3rd run at the same thing, I suggest we have a bit of a rethink and implement this a bit differently, taking on board some of the thinking from the [excellent dialogue you initiated last week on the API design](https://gds.slack.com/archives/C0776B04EJU/p1780566038307179).

Instead of this relatively complex endpoint let's implement something simpler:

`GET /api/rendered_blocks/{embedCode}` 

which returns

```json
{ 
  "embed_code": "{{embed:content_block_pension:state-pension#one_off_arrears}}",
  "title": "State Pension",
  "block_type": "Pension",
  "html": "<div class=\"content-block content-block--pension\" ...>...</div>"
}
```

**Notes**:

- takes exactly **one** embed code as a url param

### Please bear in mind
- small atomic commits telling a story
- keep the cucumber feature simple in implementation

PR #710, opened 2026-06-09
Cm 952/api endpoint to render single code

tobybatch:


engines/api/config/routes.rb
```diff
@@ -1,3 +1,3 @@
 Api::Engine.routes.draw do
-  get "blocks/search", to: "blocks#search"
+  get "blocks", to: "blocks#search"
```
edavey, line 2: It's important that sequence of commits tell the story of why and how the code is changing in this PR. Therefore this commit needs to include the clear rationale for WHY we're moving discovery from its own `/search` endpoint to the `/blocks` endpoint.

app/public/models/content_block.rb
```diff
@@ -10,6 +10,13 @@ def from_edition_id(edition_id)
       edition = Edition.find(edition_id)
       ContentBlock.new(edition)
     end
+
+    def from_embed_code(embed_code)
+      document = Document.find_by(embed_code:)
+      return nil unless document&.latest_published_edition
+
+      ContentBlock.new(document.latest_published_edition)
```
edavey: To avoid 2 queries it might be worth using a tmp, eg. something like

```rb
document = Document.find_by(embed_code:)
latest_published_edition = document&.latest_published_edition

return nil unless latest_published edition

latest_published_edition
```

spec/unit/app/public/content_block_spec.rb
```diff
@@ -33,12 +33,31 @@
   describe ".from_edition_id" do
     it "returns a content block object" do
       allow(Edition).to receive(:find).with(edition.id).and_return(edition)
-      allow(ContentBlock).to receive(:new).with(edition).and_return(content_block)
+      allow(ContentBlock).to receive(:new).with(edition).and_return(document)
```
edavey: I don't think you're intending to change the description of `ContentBlock.from_edition_id`?

spec/unit/app/public/content_block_spec.rb
```diff
@@ -33,12 +33,31 @@
...
     end
   end
 
+  describe ".from_embed_code" do
+    let(:published_edition) { build(:edition, document: document) }
+
+    it "returns a content block object for the latest published edition" do
+      allow(Document).to receive(:find_by).with(embed_code: document.embed_code).and_return(document)
+      allow(document).to receive(:latest_published_edition).and_return(published_edition)
+      allow(ContentBlock).to receive(:new).with(published_edition).and_return(content_block)
```
edavey: I think stubbing the class we're testing is a bit undermining. The example you give is very clear:

> returns a content block object for the latest published edition

and as `ContentBlock` delegates its `id` to its `Edition` we could perhaps use that attribute to establish the link between the expected "latest published edition" and the Edition which is hidden inside our `ContentBlock`? Something like:

```rb
it "returns a content block object for the latest published edition" do
  allow(Document).to receive(:find_by).with(embed_code: document.embed_code).and_return(document)
  allow(document).to receive(:latest_published_edition).and_return(published_edition)
  
  content_block = ContentBlock.from_embed_code(document.embed_code)
  expect(content_block.id).to eq(published_edition.id)
end
```

app/public/models/content_block.rb
```diff
@@ -10,6 +10,13 @@ def from_edition_id(edition_id)
       edition = Edition.find(edition_id)
       ContentBlock.new(edition)
```
edavey, line 11: Again, we need to include a fuller description of this step in the narrative we provide in our commit message. The convention is to use the present tense, so it would be something like:

> Add `ContentBlock.from_embed_code`...

See https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/working-with-git.html#commits

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -6,6 +6,18 @@ def search
     render json: Api::ResultsPresenter.present(result, request.original_url)
   end
 
+  def render_block
+    embed_code = params[:embed_code]
+    block = ContentBlock.from_embed_code(embed_code)
+    if block
+      base_embed_code_for_lookup = embed_code.gsub(/[\/#][^}]+/, "")
+      block = ContentBlock.from_embed_code(base_embed_code_for_lookup)
+      render html: block.render(embed_code)
```
edavey: Can you explain why we're calling `.from_embed_code` twice? Once with the  given embed code (which presumably may contain an "internal content path" or a "format") and once with the"base" embed code (the embed code for the block)?

Probably I'm misunderstanding but don't we wish to render the content which matches the given embed code, whether that be:

- an embed code for the default block eg. `{{embed:content_block_contact:main-office}}` 
- an embed code for internal content eg. `{{embed:content_block_contact:main-office/contact_links/new_state_pension#new_state_pension_link}}` , or
- an embed code for a format eg. `{{embed:content_block_time_period:tax-year#long_form}}"`

A good way to document this behaviour might be using cucumber scenarios e.g.

```gherkin
Scenario: Render the default block when given an embed code without internal content path or format

Scenario: Render the internal value when given an embed code which includes an internal content path

Scenario: Render the specified format when given an embed code with a format
```

engines/api/swagger/v1/swagger.json
```diff
@@ -62,7 +62,7 @@
                   "current_page": 1,
                   "links": [
                     {
-                      "href": "http://www.example.com/api/blocks/search?page=1",
+                      "href": "http://www.example.com/api/blocks?page=1",
                       "rel": "self"
```
edavey, line 66: I think this belongs with the first commit, where you change the existing discovery endpoint?

engines/api/spec/requests/api_spec.rb
```diff
@@ -264,6 +265,52 @@
     end
   end
 
+  path "/blocks/{embed_code}/render" do
```
edavey, line 268: hmm, something isn't working as planned! I was expecting to see a second `/blocks/{embedCode}/render` endpoint in our Swagger docs...

Image

edavey (commented):
👍 Looking good, thks! This is much easier to review and understand.

I've made a few small comments as there are a some remaining questions/issues...

Gemfile.lock
```diff
@@ -112,7 +112,7 @@ GEM
       debug_inspector (>= 1.2.0)
     bootsnap (1.24.6)
```
edavey, line 113: I don't think you mean to include this `Bump bootsnap from 1.24.5 to 1.24.6` commit?

app/public/models/content_block.rb
edavey, line 9: Thanks for this. Perhaps we could be a little clearer in the commit title, I found:

> Added from_embed_code to public cb model

a little cryptic. Could we be plainer (and also use the the present tense), e.g. something like:

> Add ::from_embed_code to ContentBlock

Gemfile.lock
```diff
@@ -1051,9 +1051,9 @@ CHECKSUMS
   binding_of_caller (2.0.0) sha256=e53eebf8428a85587aede7b43a83e7419eb85b8b690938a96aa330d03052d517
   block_preview (0.0.1)
   bootsnap (1.24.6) sha256=c60bab88c70332290f0a2636a288f675299eb4f804a02a3c085b42eca9da164a
-  brakeman (8.0.4) sha256=7bf921fa9638544835df9aa7b3e720a9a72c0267f34f92135955edd80d4dcf6f
+  brakeman (8.0.5) sha256=03735f9690d3fd4b32d66aacbf0a6d15a84266bdd06b32c05c8ecc8f6021d2be
```
edavey, line 1054: this does not belong here, I don't think you mean to include it?

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -8,10 +8,9 @@ def search
 
   def render_block
     embed_code = params[:embed_code]
-    block = ContentBlock.from_embed_code(embed_code)
+    base_embed_code = embed_code.gsub(/[\/#][^}]+/, "")
+    block = ContentBlock.from_embed_code(base_embed_code)
     if block
-      base_embed_code_for_lookup = embed_code.gsub(/[\/#][^}]+/, "")
-      block = ContentBlock.from_embed_code(base_embed_code_for_lookup)
       render html: block.render(embed_code)
```
edavey, line 14: I think this belongs in the earlier commit ("Add render route for blocks") where you add the `render_block` method? And the additions to the swagger spec should be part of that commit also?

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -8,10 +8,9 @@ def search
 
   def render_block
     embed_code = params[:embed_code]
-    block = ContentBlock.from_embed_code(embed_code)
+    base_embed_code = embed_code.gsub(/[\/#][^}]+/, "")
+    block = ContentBlock.from_embed_code(base_embed_code)
```
edavey, line 12: so, if I understand correctly, if my embed code includes either a "format" segment or an "internal content path", those additional specifiers are discarded and the default block is always returned. I don't see this described in any test, so I think we need to either:

- include a spec which illustrates this behaviour
- omit this `base_embed_code = embed_code.gsub(/[\/#][^}]+/, "")` refinement for now

i.e. we want all our behaviour to be described in automated tests.

edavey (commented):
Thanks for the refinements.

There are still some structural problems with this PR. How about we join forces on this and I can show you what I mean? That might be easier?

Image

PR #718, opened 2026-06-12
CM-942: Update schedule publishing page error messages

patrickjfl:
We're updating these error messages based on a content design review.


edavey (approved):
👍 thanks

PR #719, opened 2026-06-12
Fix brittle ID matching in status transition controller tests

patrickjfl:
In theory this kind of ID matching should be OK but it has caused some failures in CI, which seem to be related to clashes with existing test data. This dynamic approach should solve the problem.

### DIFF
--- a/spec/unit/app/controllers/editions/status_transitions_controller_spec.rb
+++ b/spec/unit/app/controllers/editions/status_transitions_controller_spec.rb
@@ -1,7 +1,7 @@
 RSpec.describe Editions::StatusTransitionsController, type: :controller do
   include LoginHelpers
   let(:user) { create :user }
-  let(:document) { instance_double(Document, id: 456) }
+  let(:document) { create(:document) }
   let(:edition) { create(:edition, document: create(:document)) }
   let(:transition_state_outcome) do
     {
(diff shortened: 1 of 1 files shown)


spec/unit/app/controllers/editions/status_transitions_controller_spec.rb
```diff
@@ -93,19 +93,19 @@
...
         allow(Edition).to receive(:find).and_return(edition_invalid_for_transition)
         allow(GovukError).to receive(:notify)
 
-        post :create, params: { id: 234, transition: :ready_for_review }
+        post :create, params: { id: edition_invalid_for_transition.id, transition: :ready_for_review }
       end
 
       it "redirects to the show page" do
-        expect(response).to redirect_to(document_path(456))
+        expect(response).to redirect_to(document_path(edition_invalid_for_transition.document.id))
```
edavey: would `expect(response).to redirect_to(document_path(document.id))` be clearer here

spec/unit/app/controllers/editions/status_transitions_controller_spec.rb
```diff
@@ -116,13 +116,13 @@
...
-          "`awaiting_review` for `Edition` with ID 234 "
+          "`awaiting_review` for `Edition` with ID #{edition_invalid_for_transition.id} "
 
         expect(GovukError).to have_received(:notify).with(
           expected_error_details,
           extra: {
-            edition_id: 234,
-            document_id: 456,
+            edition_id: edition_invalid_for_transition.id,
+            document_id: edition_invalid_for_transition.document.id,
```
edavey: and would

```rb
document_id: document.id,
```
be clearer here?

edavey (approved):
Thanks. I've made a couple of minor suggestions.

spec/unit/app/controllers/editions/status_transitions_controller_spec.rb
patrickjfl: yes, agreed, I've updated

spec/unit/app/controllers/editions/status_transitions_controller_spec.rb
patrickjfl: updated

PR #720, opened 2026-06-12
Validate base_path within BlockPreview to avoid SSRF attack

patrickjfl:
Until now we haven’t been validating the base_path within BlockPreview, which could in theory leave us open to a server-side request forgery attack that could expose a JWT token to an attacker. This basic defensive change should protect us from this possibility.

### DIFF
--- a/engines/block_preview/app/model/block_preview/preview_html.rb
+++ b/engines/block_preview/app/model/block_preview/preview_html.rb
@@ -9,6 +9,7 @@ module BlockPreview
     ERROR_HTML = "<html><head></head><body><p>Preview not found</p></body></html>".freeze
 
     class HtmlSnapshotError < StandardError; end
+    class UnsafePathError < StandardError; end
 
     def initialize(content_id:, block:, base_path:, locale:, state:, auth_bypass_id:)
       @content_id = content_id
@@ -30,7 +31,7 @@ module BlockPreview
       update_local_link_paths(nokogiri_html)
       update_local_form_actions(nokogiri_html, uri.scheme, uri.host)
       nokogiri_html.to_s
-    rescue HtmlSnapshotError
+    rescue HtmlSnapshotError, UnsafePathError
       ERROR_HTML
     end
 
@@ -39,7 +40,15 @@ module BlockPreview
     attr_reader :block, :content_id, :base_path, :locale, :state, :auth_bypass_id
 
     def frontend_path
-      frontend_base_path + base_path
+      clean_path = base_path.to_s.strip
+
+      invalid_sequences = ["@", "//", "\\"]
+
+      if !clean_path.start_with?("/") || invalid_sequences.any? { |seq| clean_path.include?(seq) }
+        raise UnsafePathError, "Unsafe path format"
+      end
+
+      frontend_base_path + clean_path
     end
 
     def frontend_base_path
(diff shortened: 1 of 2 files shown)


engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -39,7 +40,10 @@ def to_s
     attr_reader :block, :content_id, :base_path, :locale, :state, :auth_bypass_id
 
     def frontend_path
-      frontend_base_path + base_path
+      clean_path = base_path.to_s.strip
+      raise UnsafePathError, "Unsafe path format" if clean_path.include?("@") || clean_path.start_with?("//") || !clean_path.start_with?("/")
```
yndajas: **question:** is there a reason for not including all the suggestions in the vulnerability report?

engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -39,7 +40,10 @@ def to_s
     attr_reader :block, :content_id, :base_path, :locale, :state, :auth_bypass_id
 
     def frontend_path
-      frontend_base_path + base_path
```
yndajas, line 42: **question:** unrelated to this PR, but is `frontend_base_path` actually a base path or rather a domain/origin/host? The multiple base paths are a little confusing

engines/block_preview/spec/unit/app/models/preview_html_spec.rb
```diff
@@ -170,6 +170,73 @@
...
+        it_behaves_like "an unsafe path"
+      end
+
+      context "when it does not start with /" do
+        let(:host_base_path) { "attack.com" }
+        it_behaves_like "an unsafe path"
+      end
+
+      context "when it is an absolute URL starting with http" do
+        let(:host_base_path) { "http://attack.com" }
```
yndajas, line 234: **question:** would `"/http://attack.com"` or `"/something/http://attack.com"` be okay?

engines/block_preview/app/model/block_preview/preview_html.rb
patrickjfl: good point, I've updated to cover them all

engines/block_preview/spec/unit/app/models/preview_html_spec.rb
patrickjfl: it would have been, but shouldn't be with my updated changes

engines/block_preview/app/model/block_preview/preview_html.rb
patrickjfl: agreed, it is a domain/origin/host. it is confusing. we can look at that next week

engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -39,7 +40,15 @@ def to_s
...
+      clean_path = base_path.to_s.strip
+
+      invalid_sequences = ["@", "//", "\\"]
```
yndajas, line 45: **question:** does this need to cover single backslash? Or is the double needed for escaping?

yndajas, line 45: **suggestion:** I think you could probably use the shorthand `%w[@ // \\]` here, but if Rubocop isn't complaining 🤷🏻‍♀️

yndajas (approved):

engines/block_preview/app/model/block_preview/preview_html.rb
patrickjfl: I think the double covers the single because it's the escape character, yeah

PR #723, opened 2026-06-15
CM-983 Strengthen SSRF defences in block preview engine

edavey:
## Strengthen SSRF defences in block preview engine

[PR #720 (merged 12 Jun)](https://github.com/alphagov/content-block-manager/pull/720) added an initial fix for a reported SSRF vulnerability in the block_preview engine. That fix introduced a denylist (`@, //, \`) to reject dangerous `base_path` values before URL construction.

This PR builds on that work with structural improvements which should make the vulnerability class impossible.

1. Rename `frontend_base_path` to `frontend_origin`

2. Replace string concatenation with URI component assignment 

3. Verify trusted host before attaching credentials This ensures credentials are never sent to an untrusted host, even if earlier validation were bypassed in a future refactoring.

4. Replace `FormSubmission` domain check with explicit host allowlist

See individual commits for more detail.

### DIFF
--- a/engines/block_preview/app/model/block_preview/form_submission.rb
+++ b/engines/block_preview/app/model/block_preview/form_submission.rb
@@ -1,8 +1,10 @@
-# Represents a form submission to a gov.uk domain and its resulting behavior.
+# Represents a form submission to a GOV.UK frontend and its resulting
+# redirect behaviour. Only allows requests to known frontend hosts
+# as determined by Plek.
 #
 # @example Basic usage
 #   submission = FormSubmission.new(
-#     url: "https://example.gov.uk/form",
+#     url: "https://www.gov.uk/form",
 #     body: { field: "value" },
 #     method: "post"
 #   )
@@ -12,7 +14,7 @@ module BlockPreview
   class FormSubmission
     # @return [URI] The parsed URI of the target URL
     # @example
-    #   URI("https://example.gov.uk/form")
+    #   URI("https://www.gov.uk/form")
     attr_reader :uri
 
     # @return [Hash] The form data to be submitted
@@ -28,7 +30,7 @@ module BlockPreview
     # Raised when the HTTP response is not as expected
     class UnexpectedResponseError < StandardError; end
 
-    # Raised when the URL is not a gov.uk domain
+    # Raised when the URL host is not in the allowed frontend hosts
     class UnexpectedUrlError < StandardError; end
 
     # Initializes a new FormSubmission.
@@ -36,13 +38,14 @@ module BlockPreview
     # @param url [String] The target URL to submit the form to
     # @param body [Hash] The form data to be submitted
     # @param method [String] The HTTP method to use for the request
-    # @raise [UnexpectedUrlError] if the URL is not a gov.uk domain
+    # @raise [UnexpectedUrlError] if the URL host is not an allowed
+    #   frontend host
     #
     def initialize(url:, body:, method:)
       @uri = URI.parse(url)
       @body = body
       @method = method
-      raise UnexpectedUrlError unless uri.host&.ends_with?("gov.uk")
+      raise UnexpectedUrlError unless allowed_host?
     end
 
     # The path component of the redirect location resulting from the submission.
@@ -59,7 +62,18 @@ module BlockPreview
       location.request_uri
     end
 
+    def self.allowed_hosts
+      [
+        Plek.website_root,
+        Plek.external_url_for("draft-origin"),
+      ].map { |url| URI.parse(url).host }.compact.uniq
+    end
+
   private
+
+    def allowed_host?
+      self.class.allowed_hosts.include?(uri.host)
+    end
 
     def http_uri?(uri)
       uri.is_a?(URI::HTTP) || uri.is_a?(URI::HTTPS)
--- a/engines/block_preview/app/model/block_preview/preview_html.rb
+++ b/engines/block_preview/app/model/block_preview/preview_html.rb
@@ -96,13 +96,20 @@ module BlockPreview
     end
 
     def html_snapshot_from_frontend(uri)
+      verify_trusted_host!(uri)
       uri = add_auth_bypass_token_to_uri(uri) if draft?
-      # Net::HTTP.get_response doesn't work with Addressable::URI, so we need to convert it to a standard URI
       response = Net::HTTP.get_response(URI.parse(uri.to_s))
       if response.code == "200"
         Nokogiri::HTML.parse(response.body)
       else
         raise HtmlSnapshotError
+      end
+    end
+
+    def verify_trusted_host!(uri)
+      expected_host = Addressable::URI.parse(frontend_origin).host
+      unless uri.host == expected_host
+        raise UnsafePathError, "Untrusted host"
       end
     end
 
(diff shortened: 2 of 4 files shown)


engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -48,22 +48,23 @@ def frontend_path
         raise UnsafePathError, "Unsafe path format"
       end
 
-      frontend_base_path + clean_path
+      frontend_origin + clean_path
```
patrickjfl: One question I have from looking at this on Friday - does "frontend" here refer to GOVUK front end? Or our "front end"? Either way I think it's too vague a reference within a microservice architecture, would be good to refine it.

engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -39,16 +39,27 @@ def to_s
...
-    def frontend_path
-      clean_path = base_path.to_s.strip
+    def frontend_uri
+      validate_base_path!
+      uri = Addressable::URI.parse(frontend_origin)
+      expected_host = uri.host
+      uri.path = base_path.to_s.strip
+      uri.query = nil
+      uri.fragment = nil
+      unless uri.host == expected_host
```
patrickjfl, line 49: how would this condition ever fail, if we're setting `expected_host = uri.host` on line 45?

engines/block_preview/app/model/block_preview/preview_html.rb
```diff
@@ -96,8 +96,8 @@ def draft?
     end
 
     def html_snapshot_from_frontend(uri)
+      verify_trusted_host!(uri)
       uri = add_auth_bypass_token_to_uri(uri) if draft?
```
patrickjfl, line 100: nice

engines/block_preview/spec/unit/app/models/preview_html_spec.rb
```diff
@@ -268,6 +268,34 @@
...
+          "#{Plek.website_root}/some-path",
+        )
+        expect {
+          preview.send(:verify_trusted_host!, trusted_uri)
+        }.not_to raise_error
+      end
+
+      it "does not append the auth token when host is untrusted" do
+        untrusted_uri = Addressable::URI.parse("https://evil.com/path")
+        expect(JWT).not_to receive(:encode)
```
patrickjfl, line 342: for clarity and completeness maybe it would also be good to assert that `add_auth_bypass_token_to_uri` is not called?

engines/block_preview/app/model/block_preview/form_submission.rb
```diff
@@ -59,8 +62,19 @@ def redirect_path
       location.request_uri
     end
 
+    def self.allowed_hosts
+      [
+        Plek.website_root,
+        Plek.external_url_for("draft-origin"),
+      ].map { |url| URI.parse(url).host }.compact.uniq
```
patrickjfl, line 69: should we be using `Addressable::URI.parse` here too? (not clear on the difference between the two)

patrickjfl (approved):
Great work Ed, looks like a big improvement to me. Just a few questions and clarifications sought but nothing major.

engines/block_preview/app/model/block_preview/preview_html.rb
edavey: This refers to the [frontend app](https://github.com/alphagov/frontend). In the dev stack `frontend_origin` is `http://frontend.dev.gov.uk`, which is `development_origin` defined on line 82 as:

```rb
    # There are multiple rendering apps for GOV.UK. In non-dev environments,
    # the Router app determines the rendering app to use. We don't have access
    # to this in dev, so we need to get the rendering app from the Publishing
    # API and construct the origin that way.
    def development_origin
      @development_origin ||= begin
        publishing_api_response ||= Public::Services.publishing_api.get_content(content_id)
        Plek.external_url_for(rendering_app(publishing_api_response))
      end
    end
```

To me, "frontend" seems like a clear usage here, within publishing, where the term has a particular usage: it's  the name of the application which renders content to end users, pulling the content from Content Store (and in the future directly from Publishing API's new GraphQL API).

engines/block_preview/app/model/block_preview/preview_html.rb
edavey: At present no, this is not possible (we believe), but the idea here is to add a "belt and braces" guard in case in the future there is somehow a means of bypassing the base_path validation or mutating the uri host. I've added a commit with some illustrations:

```
BlockPreview::PreviewHtml
  path validation
    origin verification
      if input validation is somehow bypassed
        with base path set to original attack vector (@ as userinfo delimiter)
          URI component assignment prevents host hijacking
        with base path set to a protocol-relative path
          URI component assignment prevents host hijacking
        if path assignment is somehow able to mutate the URI host
          raises UnsafePathError
```

But perhaps this is all overkill? What do you think?

engines/block_preview/spec/unit/app/models/preview_html_spec.rb
edavey: In this situation I tend to think less is more. The example seems clearer by being more concise. The key assertions here are:

1. that we raise the expected `UnsafePathError`
2. that we aren't generating a JWT

If we assert that we aren't calling `add_auth_bypass_token_to_uri`:

```rb
    def add_auth_bypass_token_to_uri(uri)
      uri.query_values = (uri.query_values || {}).merge({ token: auth_bypass_token })
      uri
    end
```

we're really talking about whether or not we've called `auth_bypass_token`:

```rb
   def auth_bypass_token
      JWT.encode(
        {
          "sub" => auth_bypass_id,
          "content_id" => content_id,
          "iat" => Time.zone.now.to_i,
          "exp" => bypass_token_expiry_date.to_i,
        },
        ENV["AUTHENTICATING_PROXY_JWT_AUTH_SECRET"],
        "HS256",
      )
    end
```

and here the key assertion is whether we've called `JWT.encode` or not? 

So more assertions wouldn't be adding anything material, I think? Does that make sense?

engines/block_preview/app/model/block_preview/form_submission.rb
edavey: thanks for the question. I've added to the message on the final commit to note that:

> We use the standard library `URI.parse` here rather than `Addressable::URI` (which
> we introduced to `PreviewHtml`)  because here we are parsing
> Plek-generated origin URLs:
> 
>   - Plek.website_root,
>   - Plek.external_url_for(...))
> 
> which are trusted, well-formed strings that are always simple
> `https://hostname`/ values.
> 
> `URI.parse` is adequate for extracting the host from these.
> There's no untrusted input involved and no risk of ambiguous parsing.

engines/block_preview/app/model/block_preview/preview_html.rb
patrickjfl: i wasn't aware of this but according to Gemini:

> Because of how some URI libraries handle mutations, assigning a path that starts with // can cause the internal state of the URI object to re-parse the entire string.

so I think not overkill, a useful defensive guard.

engines/block_preview/spec/unit/app/models/preview_html_spec.rb
patrickjfl: yes, I see what you mean. I just saw that `add_auth_bypass_token_to_uri` was the next function call so my scanny brain was looking for an assertion related to that. But I think it's fine as it is, the JWT assertion is clear.

graycodes (approved):

PR #731, opened 2026-06-17
Add documentation on changing the structure of a content block

patrickjfl:
This is based on our recent experience of updating the structure of the pension block to remove the currency prefix from the “amount” field value.


tobybatch (approved):
LGTM

edavey (approved):
Overall I think this is a great start but I find it a little abstract and personally I would feel more confident with a guide which steps through the process with more concrete examples. As discussed offline, perhaps we can come up with a flow chart to illustrate? That might help us identify possible improvements in how the E2E tests are orchestrated? 

I think there are two key questions which we'd like to be able to answer confidently when changing blocks: 

1. **Has the gem been adapted to temporarily handle both new and old formats safely?** Rendering the block internally: this is done by the "tools" gem in CBM itself and in publishing apps such as Whitehall and "mainstream" Publisher. 

2. **Do we know how the CBM UX (and therefore the E2E test) will behave at each step in the block migration process?** Editing the block in CBM: we saw how our time period's new format caused existing values to become "invalid" until the data migration was performed. Being clear on the timing/impact of the data migration is important for understanding the user experience and for managing the release process which depends on E2E tests passing.

PR #734, opened 2026-06-18
Update BUNDLED WITH to 4.0.14 to fix lockfile checksums

edavey:
We've been seeing candidate PRs which include unexpected and erroneous changes to the Gemfile.lock checksums. Specifically the removal of:

```
bundler (4.0.14) sha256=d09a0a965cf772266a7e49e83610be7c2f4e49e61134c42a56804bb383cc24b8
```

This is because Dependabot runs with the latest Bundler (currently 4.0.14) and Bundler 4.x writes a SHA256 checksum for itself into the CHECKSUMS section.

If a developer runs `bundle install` locally with an older Bundler (4.0.10), it cannot verify the checksum for 4.0.14 and removes the line.

Aligning BUNDLED WITH to the current Bundler version prevents this issue.

In our local dev environments we need to run `gem install bundler` periodically to stay in step with the version Dependabot uses.


patrickjfl (approved):
Thanks for looking into this!

PR #735, opened 2026-06-18
Dependabot updates to address alerts: 95 & 94

edavey:
This addresses two security updates where dependabot is unable to resolve dependencies. See:

1. https://github.com/alphagov/content-block-manager/security/dependabot/95
2. https://github.com/alphagov/content-block-manager/security/dependabot/94


patrickjfl (approved):
thanks Ed

PR #738, opened 2026-06-19
Fix API query

patrickjfl:
We have had a flaky query spec unit test that is symptomatic of an error in our query implementation, and has been sporadically blocking our CI pipeline. Until now the chaining of  `.merge(Edition.most_recent_for_document).merge(Edition.published)` has meant that a Document is excluded entirely from the results if its most recent Edition is not published (i.e. is in draft).

To address this I’ve added a new method on the Edition class to get the most recent published document, and ensured that in the results method we are returning the latest *published* edition, rather than the most recent edition.

### DIFF
--- a/app/public/queries/content_block/query.rb
+++ b/app/public/queries/content_block/query.rb
@@ -13,7 +13,7 @@ class ContentBlock
 
     def results
       Result.new(
-        blocks: paginated_results.map { |document| ContentBlock.new(document.most_recent_edition) },
+        blocks: paginated_results.map { |document| ContentBlock.new(document.latest_published_edition) },
         current_page: paginated_results.current_page,
         total_pages: paginated_results.total_pages,
         total_count: paginated_results.total_count,
@@ -34,8 +34,8 @@ class ContentBlock
 
     def unpaginated_results
       Document.joins(:editions)
-              .merge(Edition.most_recent_for_document)
               .merge(Edition.published)
+              .merge(Edition.most_recent_published_for_document)
               .extending(Scopes)
               .by_block_type(filters[:block_type])
               .by_lead_organisation_id(filters[:lead_organisation_id])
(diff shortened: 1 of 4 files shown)


tobybatch (approved):

graycodes:
I'm confused about this. I can definitely see documents whose latest edition is a draft in integration / staging but you've written this:

> We have had a flaky query spec unit test that is symptomatic of an error in our query implementation. Until now the chaining of  `.merge(Edition.most_recent_for_document).merge(Edition.published)` has meant that a Document is excluded entirely from the results if its most recent Edition is not published (i.e. is in draft).

app/public/queries/content_block/query.rb
```diff
@@ -34,8 +34,8 @@ def page
 
     def unpaginated_results
       Document.joins(:editions)
-              .merge(Edition.most_recent_for_document)
               .merge(Edition.published)
```
edavey, line 37: do you need to have:

```rb
.merge(Edition.published)
```

as well as

```rb
 .merge(Edition.most_recent_published_for_document)
```

here?

patrickjfl:
> I'm confused about this. I can definitely see documents whose latest edition is a draft in integration / staging but you've written this:
> 
> > We have had a flaky query spec unit test that is symptomatic of an error in our query implementation. Until now the chaining of  `.merge(Edition.most_recent_for_document).merge(Edition.published)` has meant that a Document is excluded entirely from the results if its most recent Edition is not published (i.e. is in draft).

so the `/blocks/search` API endpoint in staging/integration was returning the draft blocks?

app/public/queries/content_block/query.rb
patrickjfl: good point, I think i can get rid of `.merge(Edition.published)` can't i?

graycodes:
> so the `/blocks/search` API endpoint in staging/integration was returning the draft blocks?

Oh you know what, I'm being slow. I was thinking of the search endpoint of the main app, not the API

PR #739, opened 2026-06-19
Add render end point

tobybatch:


app/public/models/content_block.rb
```diff
@@ -10,6 +10,18 @@ def from_edition_id(edition_id)
       edition = Edition.find(edition_id)
       ContentBlock.new(edition)
```
edavey, line 11: Nice work! Just atiny nit-pick on the commit message: as we have a convention to always try to accompany changes to the source with changes to the test suite, the section describing the changes to the suite is not needed. We're going to look at that part of the commit first to understand the changes.

app/public/models/content_block.rb
```diff
@@ -10,6 +10,18 @@ def from_edition_id(edition_id)
...
     end
+
+    def from_embed_code(embed_code)
+      embed_code = embed_code.to_s
+      return nil if embed_code.empty?
```
edavey, line 16: could we replace this with a simpler guard e.g.

```rb
return if embed_code.blank?
```

I'm not sure what scenario the `.to_s` is handling? 

If you think `blank?` is the right guard here, we might want to an example to describe how given an empty string we:

```rb
expect(ContentBlock.from_embed_code("")).to be_nil
```

spec/unit/app/public/content_block_spec.rb
```diff
@@ -39,6 +39,53 @@
...
+      expect(content_block.id).to eq(published_edition.id)
+    end
+
+    it "returns a content block object using an embed code with a path" do
+      expect(valid_field_path).to be_present
+
+      embed_code_with_path = document.embed_code_for_field(valid_field_path)
+
+      allow(Document).to receive(:find_by).with(embed_code: document.embed_code).and_return(document)
+      allow(document).to receive(:latest_published_edition).and_return(published_edition)
```
edavey: A couple of structural suggestions. RSpec allows us to organise our examples quite flexibly. 

### 1. Consider using contexts for setting up scenarios

The description of `from_embed_code` currently has this shape:

```
.from_embed_code
  returns a content block object using the base embed code
  returns a content block object using an embed code with a format
  returns a content block object using an embed code with a path
  handles a null embed code gracefully
```

If we used "a context" for each of the different scenarios we're describing we could give it a shape like:

```
.from_embed_code
  when given the base embed code
    returns a content block wrapping the published edition of the matching document
  when the embed code includes a format segment
    returns a content block wrapping the published edition of the matching document
  when the embed code includes an internal content path segment
    returns a content block wrapping the published edition of the matching document
  when no embed code is given
    handles a null embed code gracefully  
```

which might be easier to parse visually?

### 2. Consider extracting common stubs to reduce visual noise

We're repeating 

```rb
allow(Document).to receive(:find_by).with(embed_code: document.embed_code).and_return(document)
allow(document).to receive(:latest_published_edition).and_return(published_edition)
```
 in each example.

We could declare these stubs once in a "before" block at the start of the "description" between the individual "contexts", something like:

```rb
describe ".from_embed_code" do
  let(:document) do
  let(:published_edition) { build(:edition, document: document) }

  before do
    allow(Document).to receive(:find_by).with(embed_code: document.embed_code).and_return(document)
    allow(document).to receive(:latest_published_edition).and_return(published_edition)
  end

  context "when given the base embed code" do
    it "returns a content block wrapping the published edition of the matching document" do
  end

  context "when the embed code includes a format segment" do
    it "returns a content block wrapping the published edition of the matching document" do
  end

  context "when the embed code includes an internal content path segment" do
    it "returns a content block wrapping the published edition of the matching document" do
  end

  context "when no embed code is given" do
    it "handles a null embed code gracefully" do
  end
end
```

spec/unit/app/public/content_block_spec.rb
```diff
@@ -39,6 +39,53 @@
...
+
+      content_block = ContentBlock.from_embed_code(document.embed_code)
+      expect(content_block.id).to eq(published_edition.id)
+    end
+
+    it "returns a content block object using an embed code with a format" do
+      expect(valid_format).to be_present
+
+      embed_code_with_format = document.embed_code_for_format(valid_format)
```
edavey: do you think this might make for clearer documentation if we were to use "literal" values for the embed code variants which include the "format" or "internal content path" segments? Personally I think that if I was trying to understand what's going on here I'd find it more helpful to see something like:

```rb
    context "when the embed code includes a format segment" do
      it "returns a content block wrapping the published edition of the matching document" do
        content_block = ContentBlock.from_embed_code(
          "{{embed:content_block_time_period:tax-year#long_form}}"
        )
        expect(content_block.id).to eq(published_edition.id)
      end
    end
```

engines/api/app/controllers/api/blocks_controller.rb
```diff
@@ -6,6 +6,16 @@ def search
     render json: Api::ResultsPresenter.present(result, request.original_url)
   end
```
edavey, line 7: Nice! 

But I think we can leave out the following from the commit message as this test document is taken for granted:

```
- add API acceptance coverage for base embed codes, unknown embed codes,
  internal content paths, and format-specific embed codes
- add request/spec coverage for the render endpoint and expected response types
- update Swagger docs to describe the new endpoint and responses
```

engines/api/features/api/step_definitions/api_steps.rb
```diff
@@ -60,3 +62,62 @@
...
+  visit "/api/blocks/#{encoded_embed_code}/render"
+end
+
+Then("the response is rendered HTML") do
+  expect(page.status_code).to eq(200)
+  expect(page.response_headers["Content-Type"]).to include("text/html")
+  expect(page.source).to include("content-block")
+end
+
+Then("the response contains rendered content for {string}") do |title|
```
edavey: Currently we're asserting that the response includes a string such as `"embed:content_block_time_period:"`.

Looking at the response:
    
  ```html
  <div 
    class="content-block content-block--time_period"
    data-content-block=""
    data-document-type="time_period"
    data-content-id="366ad8c5-d717-420e-8063-e38447bcec80"
    data-embed-code="{{embed:content_block_time_period:factory-example-title--2}}"
>
  </div>
  ```
  
  I see that the assertion is really about the composition of the `data-embed-code` attribute. So perhaps we could be clearer about what we're describing here. e.g. by naming the step something like
  
 ```rb
Then("the response contains a data-embed-code attribute with the fragment {string}") |embed_code_fragment|
 ```

engines/api/spec/requests/api_spec.rb
```diff
@@ -264,6 +265,99 @@
...
+        end
+
+        run_test! do |response|
+          expect(response.content_type).to include("text/html")
+          expect(response.body).to include("content-block")
+          expect(response.body).to include("State Pension")
+        end
+      end
+
+      response "200", "renders HTML for a field-path embed code (embed code containing a slash)", document: false do
```
edavey: I think it will be better to stick to the terminology of "internal content path" here e.g.

> renders the specified sub content for an internal content path

engines/api/spec/requests/api_spec.rb
```diff
@@ -264,6 +265,99 @@
...
+            title: "State Pension Field",
+            lead_organisation_id: SecureRandom.uuid,
+            document: @document,
+          )
+        end
+
+        let(:embed_code) { CGI.escape(@document.embed_code_for_field("rates/rate1/amount")) }
+
+        run_test! do |response|
+          expect(response.content_type).to include("text/html")
```
edavey, line 387: I feel like perhaps we need to go rather further here to be explicit and helpful in our description of the expected response. I believe it's important to use a realistic example also. By:

- including a `details` hash in our Edition which matches the internal content path we use we'll render renders the expected `amount`
- using a clear and literal description of the response give the reader an example payload, inline.

Something like:

```rb
    response "200", "renders HTML for a field-path embed code (embed code containing a slash)", document: false do
        before do
          @document = create(
            :document,
            block_type: "pension",
            sluggable_string: "state-pension",
            content_id: "11111111-2222-4333-8444-666666666666",
          )
          create(
            :edition,
            :published,
            details: {
              "rates" => {
                "weekly-rate" => {
                  "title" => "Weekly rate",
                  "amount" => "999.69",
                  "frequency" => "a week"
                }
              }
            },
            title: "State Pension",
            lead_organisation_id: SecureRandom.uuid,
            document: @document,
          )
        end

        let(:embed_code) { CGI.escape(@document.embed_code_for_field("rates/weekly-rate/amount")) }

        let(:expected_response) do
          <<~HTML
            <span
              class="content-block content-block--pension"
              data-content-block=""
              data-document-type="pension"
              data-content-id="11111111-2222-4333-8444-666666666666"
              data-embed-code="{{embed:content_block_pension:state-pension/rates/weekly-rate/amount}}"
              >
              £999.69
            </span>
          HTML
        end

        run_test! do |response|
          expect(response.content_type).to include("text/html")
          expect(
            normalise_html(response.body)).to eq(
            normalise_html(expected_response)
          )
        end
      end

       response "200", "renders HTML for a field-path embed code (embed code containing a slash)", document: false do
        before do
          @document = create(
            :document,
            block_type: "pension",
            sluggable_string: "state-pension",
            content_id: "11111111-2222-4333-8444-666666666666",
          )
          create(
            :edition,
            :published,
            details: {
              "rates" => {
                "weekly-rate" => {
                  "title" => "Weekly rate",
                  "amount" => "999.69",
                  "frequency" => "a week"
                }
              }
            },
            title: "State Pension",
            lead_organisation_id: SecureRandom.uuid,
            document: @document,
          )
        end

        let(:embed_code) { CGI.escape(@document.embed_code_for_field("rates/weekly-rate/amount")) }

        let(:expected_response) do
          <<~HTML
            <span
              class="content-block content-block--pension"
              data-content-block=""
              data-document-type="pension"
              data-content-id="11111111-2222-4333-8444-666666666666"
              data-embed-code="{{embed:content_block_pension:state-pension/rates/weekly-rate/amount}}"
              >
              £999.69
            </span>
          HTML
        end

        run_test! do |response|
          expect(response.content_type).to include("text/html")
          expect(
            normalise_html(response.body)).to eq(
            normalise_html(expected_response)
          )
        end
      end

   def normalise_html(html)
    fragment = Nokogiri::HTML.fragment(html)
    fragment.traverse do |node|
      if node.text?
        node.content = node.content.squish
        node.remove if node.content.empty?
      end
    end
    fragment.to_html
  end
```

Do you think that a more realistic example, with the response presented "literally" makes for clearer documentation and is worth the extra effort?

engines/api/spec/requests/api_spec.rb
```diff
@@ -264,6 +265,99 @@
...
+        end
+      end
+
```
edavey, line 451: I think we able need an example which shows the "format" example? e.g. something like:

```rb
    response "200", "renders HTML for the given format", document: false do
        before do
          @document = create(
            :document,
            block_type: "time_period",
            sluggable_string: "tax-year",
            content_id: "11111111-2222-4333-8444-666666666669",
          )
          create(
            :edition,
            :published,
            details:  {
              "date_range" => {
                "start" => "2026-04-06T00:00:00+01:00",
                "end" => "2027-04-05T23:59:00+01:00"
              }
            },
            title: "Tax year",
            lead_organisation_id: SecureRandom.uuid,
            document: @document,
          )
        end

        let(:embed_code) { CGI.escape(@document.embed_code_for_format("years"))  }

        let(:expected_response) do
          <<~HTML
            <div 
              class="content-block content-block--time_period" 
              data-content-block="" 
              data-document-type="time_period" 
              data-content-id="11111111-2222-4333-8444-666666666669" data-embed-code="{{embed:content_block_time_period:tax-year#years}}"
            >
              <p class="govuk-body">
                2026-2027
              </p>
            </div> 
          HTML
        end

        run_test! do |response|
          expect(response.content_type).to include("text/html")
          expect(
            normalise_html(response.body)).to eq(
            normalise_html(expected_response)
          )
        end
      end
```

edavey (commented):
Thanks Toby, looking good! 

I've made a few further suggestions where I think we can be a bit clearer...

app/public/models/content_block.rb
tobybatch: If we get a non string back from the URL params then wouldn't it throw a runtime error on .blank?

The .to_s ensures we have a string when executing the string specific actions.

spec/unit/app/public/content_block_spec.rb
tobybatch: I agree literal values are a bit easier to read in isolation. I went with generated values deliberately so the spec stays aligned with real schema-backed embed codes and doesn’t duplicate the embed-code grammar in another place. We already have literal expectations for the exact string shape in Document specs, so here I was aiming to test that ContentBlock.from_embed_code accepts real valid variants rather than documenting the format itself. If helpful, I can make that clearer with variable naming or a short comment, but i thought we didn';t like inline code comments.

app/public/models/content_block.rb
tobybatch: Here is a link to the updated commit with the justification in it

https://github.com/alphagov/content-block-manager/pull/739/changes/075b303d0d34439d40999665255f333fd13dedb5

Gemfile.lock
```diff
@@ -1053,7 +1057,6 @@ CHECKSUMS
   bootsnap (1.24.6) sha256=c60bab88c70332290f0a2636a288f675299eb4f804a02a3c085b42eca9da164a
   brakeman (8.0.5) sha256=03735f9690d3fd4b32d66aacbf0a6d15a84266bdd06b32c05c8ecc8f6021d2be
   builder (3.3.0) sha256=497918d2f9dca528fdca4b88d84e4ef4387256d984b8154e9d5d3fe5a9c8835f
-  bundler (4.0.14) sha256=d09a0a965cf772266a7e49e83610be7c2f4e49e61134c42a56804bb383cc24b8
```
edavey, line 1056: I think you added this by mistake? Please see https://github.com/alphagov/content-block-manager/pull/734 for how we can update our local `bundler` installation to prevent this re-occurring.

app/public/models/content_block.rb
edavey: ok, thks, makes sense! So let's describe this `to_s` behaviour with an additional example then? e.g.

```rb
    context "when an integer is given" do
      before do
        allow(Document).to receive(:find_by).with(embed_code: "666").and_return(nil)
      end

      it "returns nil" do
        expect(ContentBlock.from_embed_code(666)).to be_nil
      end
    end

```


i.e. if we think it's a scenario worth covering in our implementation, then it's probably behaviour which is also worth describing in our automated tests?

engines/api/spec/requests/api_spec.rb
```diff
@@ -264,6 +265,134 @@
...
+        run_test! do |response|
+          data = JSON.parse(response.body)
+
+          expect(data).to eq({
+            "error" => "Content block not found for embed code: {{embed:content_block_pension:missing-block}}",
+          })
+        end
+      end
+
+      response "404", "returns an error when the embed code with a slash field path is unknown", document: false do
```
edavey: in the interests of consistency is it worth using the term "internal content path" here rather "slash field path"?

engines/api/spec/requests/api_spec.rb
```diff
@@ -264,6 +265,134 @@
...
+        let(:embed_code) { CGI.escape("{{embed:content_block_pension:missing-block/rates/rate1/amount}}}") }
+
+        run_test! do |response|
+          data = JSON.parse(response.body)
+          expect(data["error"]).to be_present
+        end
+      end
+    end
+  end
+
```
edavey, line 474: I think we're missing an example for the scenario where the embed code includes a "format"? e.g. something like:

```rb
response "200", "renders HTML for an embed code containing a format" do
  before do
    @document = create(
      :document,
      block_type: "time_period",
      sluggable_string: "tax-year",
      content_id: "11111111-2222-4333-8444-666666666669",
    )
    create(
      :edition,
      :published,
      details:  {
        "date_range" => {
          "start" => "2026-04-06T00:00:00+01:00",
          "end" => "2027-04-05T23:59:00+01:00"
        }
      },
      title: "Tax year",
      lead_organisation_id: SecureRandom.uuid,
      document: @document,
    )
  end

  let(:embed_code) { CGI.escape(@document.embed_code_for_format("years"))  }

  let(:expected_response) do
    <<~HTML
      <div 
        class="content-block content-block--time_period" 
        data-content-block="" 
        data-document-type="time_period" 
        data-content-id="11111111-2222-4333-8444-666666666669" data-embed-code="{{embed:content_block_time_period:tax-year#years}}"
      >
        <p class="govuk-body">
          2026-2027
        </p>
      </div> 
    HTML
  end

  run_test! do |response|
    expect(response.content_type).to include("text/html")
    expect(
      normalise_html(response.body)).to eq(
      normalise_html(expected_response)
    )
  end
end
```

edavey: I was having a little play to see how we can show these 3 different scenarios:

1. base embed code renders default block
2. internal content path renders targeted sub-content
3. format renders expected presentation

in our generated OpenAPI spec. It seems that we can get documentation for multiple scenarios of a single endpoint like:

Image

If we:

- nest the `response 200` blocks
- "deep merge" each example in an `after` block

e.g. this structure:

```
path "/blocks/{embed_code}/render" do
  get "Render a content block" do
    response "200", "renders HTML for a base embed code" do
      after do |example|
          content = example.metadata[:response][:content] || {}
          example.metadata[:response][:content] = content.deep_merge(
            "text/html" => {
              examples: {
                base_embed_code: {
                  summary: "Base embed code renders the block title",
                  value: response.body,
                },
              },
            },
          )
        end

        run_test!

    response "200", "renders HTML for an embed code containing a format" do
        before do
          # set up TimePeriod with date range
        end

        let(:embed_code) { CGI.escape(@document.embed_code_for_format("years")) }

        let(:expected_response) do


        after do |example|
          content = example.metadata[:response][:content] || {}
          example.metadata[:response][:content] = content.deep_merge(
            "text/html" => {
              examples: {
                format_segment: {
                  summary: "Format segment renders a formatted representation (e.g. tax year as years)",
                  value: response.body,
                },
              },
            },
          )
        end

        run_test! do |response|

    response "200", "renders HTML for embed code with internal content path" do
        before do
          # setup Pension with weekly rate and amount
        end

        let(:embed_code) { CGI.escape(@document.embed_code_for_field("rates/weekly-rate/amount")) }

        let(:expected_response) do

        after do |example|
          content = example.metadata[:response][:content] || {}
          example.metadata[:response][:content] = content.deep_merge(
            "text/html" => {
              examples: {
                sub_field_path: {
                  summary: "Internal content path renders the specific value (e.g. weekly rate amount)",
                  value: response.body,
                },
              },
            },
          )
        end

        run_test! do |response|
```

Gemfile.lock
edavey, line 1056: I don't think you meant to include this? Please see https://github.com/alphagov/content-block-manager/pull/734 for how we can update our local `bundler` so that these don't keep creeping in when we run `bundle install` locally...

edavey (commented):
Nice work Toby!

Just some final tiny suggestions, to:

- add a unit test for the `to_s` guard on `ContentBlock.from_embed_code`
- add an api spec for the "format" use case
- tweak the structure of the api spec to give example output for the 3 scenarios in our OpenAPI spec

Very happy to pair on this as I appreciate my requests have been a bit endless!

edavey (approved):
👍 nice work, thanks for being so patient!

PR #762, opened 2026-07-02
Removes pagination from /api/blocks search

tobybatch:
Return the full block search response directly from ContentBlock::Query instead of wrapping it in paginated API results.

This removes:
- page parameter validation
- pagination metadata and links
- dependence on absolute next-page URLs that break behind reverse proxies
- updates the presenter to reove metadata

It also simplifies clients by eliminating the need to loop through paginated responses to collect all blocks.

Update API feature coverage and step definitions to assert the new response shape (`blocks`) and remove pagination-specific scenarios.


edavey (approved):
👍 nice work, 430 lines less to maintain! Thanks for pushing this simplification forward. 🙇

PR #772, opened 2026-07-07
(CM-947) ActiveRecord Refactoring ADR

graycodes:
ADR for move to storing blocks as ActiveRecord models.

Based off the existing experimental work done to determine if this was a viable and helpful change to our approach, this ADR aims to encapsulate what we discovered. Much of the content of this ADR references the original experiment ADR.

The original ADR contains more detail in some areas where it didn't feel necessary to re-state here.

Original experiment:
https://github.com/alphagov/content-block-manager/pull/553


docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md
```diff
@@ -0,0 +1,147 @@
...
+`Block::TaxEdition`.
+
+#### 3. Content in Specialized Tables
+
+Each edition type has its own dedicated content tables with typed columns. For example a `Block::ContactEdition` may
+have a `Block::ContactEmail` and multiple `Block::ContactTelephoneNumber`.
+
+#### 4. No Generic Details Column
+
+Instead of a `details` JSON column each edition subclass can implement a `#details` method that serializes from its
```
edavey: would it perhaps be clearer to say:

> each edition subclass *will* implement a `#details` method

to make it clear that the business of representing itself in JSON is going to be a responsibility of each specialisation of block?

docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md
```diff
@@ -0,0 +1,147 @@
...
+          "properties": {
+             "description": {
+               "x-component-name": "textarea",
+               "x-govspeak-enabled": true
+               ...
+```
+- Block-specific changes should be faster to implement because they no longer need to fit into a heavily generic
+  abstraction.
+- Validation should become more explicit.
+- Future maintenance should be less risky because behaviour is easier to trace through conventional Rails patterns.
```
edavey, line 224: Is it worth also asserting that we believe that this new architecture will allows us to change develop our service (through the iterative method) faster and thereby meet users' needs sooner?

docs/architecture/decisions/experiments/0001-experiment-with-traditional-activerecord-models.md
```diff
@@ -0,0 +1,265 @@
...
+Instead of a `details` JSON column:
+
+- Each edition subclass implements a `#details` method that serializes from its associated content models
+- Example: `TimePeriodEdition#details` composes from `description` + `date_range.to_details`
+- Rationale: One-way serialization for API consumption; editing happens via associations
+
+
+#### 6. Two-Controller Pattern for Multi-Step Forms
+
+Rather than use general-purpose multi-step "wizard" code (`Workflow`), it may be preferable, especially in the case of simple blocks (e.g. `TimePeriod`) to use separate controllers for different form steps:
```
edavey: We might perhaps be clearer here that this is about making the flow control explicit rather than derived from some other more abstract mechanism?

e.g would something maybe:

> ... to use separate `TimePeriod` controllers to orchestrate the flow between the forms used to build the complete block

docs/architecture/decisions/experiments/0001-experiment-with-traditional-activerecord-models.md
```diff
@@ -0,0 +1,265 @@
...
+
+- `TimePeriodEditionsController` - Step 1: common fields (title, description, organisation)
+- `TimePeriodDateRangesController` - Step 2: type-specific fields (start/end dates)
+
+#### 7. Reuse of Existing Concerns
+
+Where possible, this experiment will aim to reuse existing behaviour which is packaged up in modules, e.g.:
+
+- `Block::Edition` includes `::Edition::HasLeadOrganisation` for organisation validation
+- `Block::TimePeriodDateRange` includes `DateValidation` for handling invalid multiparameter dates
```
edavey: I think maybe `DateValidation` isn't a good example of concerns we expect to reuse? As you say lower down, it actually belongs in the category of "behaviours" which we're confident we'll be able to implement in a simpler way?

docs/architecture/decisions/experiments/0001-experiment-with-traditional-activerecord-models.md
```diff
@@ -0,0 +1,265 @@
...
+
+Specifically, we want to understand:
+
+2. **Developer Experience:** Is it easier and faster to work with explicit Rails models and associations vs. generic JSON columns? Our experience when adding a new block type is that there are numerous differences or additional features which are needed. At face value they appear minor but in reality prove complex to introduce to an application model which is entirely generic. Our small number of classes can do everything, but at what price in terms of complexity and cognitive load.
+3. **Maintainability:** Are dedicated models with typed columns easier to understand and modify than schema definitions?
+4. **Adding Fields:** Is it simpler to add a migration for a new column vs. updating a JSON schema?
+5. **Performance:** Do typed columns with proper indexes perform better than querying within JSON?
+6. **Testing:** Are traditional model tests more straightforward than testing schema validation?
+
```
edavey, line 35: Within this "Context" section, would it be helpful to have a visual aid? In my https://github.com/alphagov/content-block-manager/pull/553 PR narrative I included a couple of diagrams which attempted to illustrate the difference between:

1. current state: complexity concentrated into few entities
2. target state: complexity spread over numerous entities

Image

and

Image

docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md
```diff
@@ -0,0 +1,185 @@
...
+None of this work is attempting to change the values stored for our blocks and this change should be invisible to the
+end users.
+
+## Consequences
+
+### Benefits
+
+- The codebase should be easier to reason about because the shape of each block type is expressed directly in models and
+  tables. We currently rely on schemas to denote a lot of behaviour in the system and the code that operates on this can
+  be hard to understand. We currentlu se a lot of custom schema properties to define behaviour e.g.
```
edavey: tiny typo:

> currentlu se

edavey (dismissed):
This is really clear, thanks.

If you feel it's helpful to leave this open for a while, whilst you work through the `TimePeriod` work, in order to make refinements / adjustments, then I think you should consider doing that, rather than feeling like you need to finalise all the details immediately.

docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md
graycodes: doh

docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md
graycodes: Yep excellent

docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md
graycodes: Ah yes good point!

docs/architecture/decisions/experiments/0001-experiment-with-traditional-activerecord-models.md
graycodes: I have just added some simple diagrams to hopefully describe these changes in two separate ways in the new ADR doc. I've now added your existing ones to this doc too. Let me know if you think there's more I could add to make things clearer.

docs/architecture/decisions/experiments/0001-experiment-with-traditional-activerecord-models.md
graycodes: 🤔 I had just copied this file exactly from your experiment branch so I was just going to commit it as-is so that we can reference it. Do you think I should update it?

edavey: yes I think we can tweak to improve!

edavey (approved):
👍  great work, thks. Please go ahead and merge when you're ready.

PR #777, opened 2026-07-10
(CM-949) Initial ActiveRecord models

graycodes:
# Description

Includes the first pass at the models for the ActiveRecord refactoring, including:

- Block::Document
- Block::Edition
- Block::TimePeriodEdition
- Block::TimePeriodDateRange

Note: This doesn't yet provide any way of interacting with these models. Follow-on work will need to be done to include the forms, controllers, etc that will allow interactions with these models. This is merely a base layer to build on. As this is fairly foundational work in this process I think it's important to split this out on its own to ensure we're all happy with it before we move on to the next phase.

Based very heavily on the work by @edavey on https://github.com/alphagov/content-block-manager/pull/553 with only minor tweaks and relating to the ADR in #772 .

### DIFF
--- a/app/models/block/document.rb
+++ b/app/models/block/document.rb
@@ -3,7 +3,18 @@ module Block
     extend FriendlyId
     friendly_id :sluggable_string, use: :slugged, slug_column: :content_id_alias, routes: :default
 
-    has_many :editions, class_name: "Block::Edition", foreign_key: :block_document_id, dependent: :destroy
+    has_many :editions,
+             class_name: "Block::Edition",
+             foreign_key: :block_document_id,
+             dependent: :destroy,
+             inverse_of: :document
+
+    has_many :time_period_editions,
+             -> { where(type: "Block::TimePeriodEdition") },
+             class_name: "Block::TimePeriodEdition",
+             foreign_key: :block_document_id,
+             dependent: :destroy,
+             inverse_of: :document
 
     before_validation :generate_content_id, on: :create
     after_validation :set_content_id_alias_and_embed_code, on: :create
--- a/app/models/block/document.rb
+++ b/app/models/block/document.rb
@@ -1,0 +1,30 @@
+module Block
+  class Document < ApplicationRecord
+    extend FriendlyId
+    friendly_id :sluggable_string, use: :slugged, slug_column: :content_id_alias, routes: :default
+
+    has_many :editions, class_name: "Block::Edition", foreign_key: :block_document_id, dependent: :destroy
+
+    before_validation :generate_content_id, on: :create
+    after_validation :set_content_id_alias_and_embed_code, on: :create
+
+    def built_embed_code
+      "{{embed:content_block_#{block_type}:#{content_id_alias}}}"
+    end
+
+    def embed_code_for_field(field_path)
+      "{{embed:content_block_#{block_type}:#{content_id_alias}/#{field_path}}}"
+    end
+
+  private
+
+    def generate_content_id
+      self.content_id ||= SecureRandom.uuid
+    end
+
+    def set_content_id_alias_and_embed_code
+      self.content_id_alias = friendly_id
+      self.embed_code = built_embed_code
+    end
+  end
+end
--- a/app/models/block/document.rb
+++ b/app/models/block/document.rb
@@ -8,6 +8,10 @@ module Block
     before_validation :generate_content_id, on: :create
     after_validation :set_content_id_alias_and_embed_code, on: :create
 
+    def title
+      editions.order(created_at: :desc).first&.title
+    end
+
     def built_embed_code
       "{{embed:content_block_#{block_type}:#{content_id_alias}}}"
     end
--- a/app/models/block/edition.rb
+++ b/app/models/block/edition.rb
@@ -1,0 +1,15 @@
+module Block
+  class Edition < ApplicationRecord
+    include ::Edition::HasLeadOrganisation
+
+    belongs_to :document, class_name: "Block::Document", foreign_key: :block_document_id
+
+    validates :title, presence: true
+
+    # Abstract method to be implemented by subclasses
+    # Returns a hash representation of the edition's details
+    def to_details
+      raise NotImplementedError, "Subclasses must implement #to_details method"
+    end
+  end
+end
--- a/app/models/block/edition.rb
+++ b/app/models/block/edition.rb
@@ -2,14 +2,31 @@ module Block
   class Edition < ApplicationRecord
     include ::Edition::HasLeadOrganisation
 
-    belongs_to :document, class_name: "Block::Document", foreign_key: :block_document_id
+    belongs_to :document, class_name: "Block::Document", foreign_key: :block_document_id, inverse_of: :editions
 
     validates :title, presence: true
+    validate :title_contains_alphanumeric_chars
+
+    before_validation :set_document_sluggable_string, on: :create
 
     # Abstract method to be implemented by subclasses
     # Returns a hash representation of the edition's details
     def to_details
       raise NotImplementedError, "Subclasses must implement #to_details method"
+    end
+
+  private
+
+    def set_document_sluggable_string
+      return unless document.present? && document.new_record?
+
+      document.sluggable_string = title if document.sluggable_string.blank?
+    end
+
+    def title_contains_alphanumeric_chars
+      if title.present? && title !~ /[a-z0-9]+/i
+        errors.add(:title, :alphanumeric)
+      end
     end
   end
 end
--- a/app/models/block/time_period_date_range.rb
+++ b/app/models/block/time_period_date_range.rb
@@ -2,6 +2,9 @@ module Block
   class TimePeriodDateRange < ApplicationRecord
     belongs_to :edition, class_name: "Block::TimePeriodEdition"
 
+    include DateValidation
+    date_attributes :start, :end
+
     validates :start, presence: true
     validates :end, presence: true
     validate :end_date_after_start_date
--- a/app/models/block/time_period_date_range.rb
+++ b/app/models/block/time_period_date_range.rb
@@ -1,0 +1,26 @@
+module Block
+  class TimePeriodDateRange < ApplicationRecord
+    belongs_to :edition, class_name: "Block::TimePeriodEdition"
+
+    validates :start, presence: true
+    validates :end, presence: true
+    validate :end_date_after_start_date
+
+    def to_details
+      {
+        "start" => start,
+        "end" => self.end,
+      }
+    end
+
+  private
+
+    def end_date_after_start_date
+      return if self.end.blank? || start.blank?
+
+      if self.end <= start
+        errors.add(:end, "must be after start date")
+      end
+    end
+  end
+end
--- a/app/models/block/time_period_edition.rb
+++ b/app/models/block/time_period_edition.rb
@@ -1,0 +1,4 @@
+module Block
+  class TimePeriodEdition < Edition
+  end
+end
--- a/app/models/block/time_period_edition.rb
+++ b/app/models/block/time_period_edition.rb
@@ -1,4 +1,7 @@
 module Block
   class TimePeriodEdition < Edition
+    has_one :date_range, class_name: "Block::TimePeriodDateRange", foreign_key: :edition_id, dependent: :destroy
+
+    accepts_nested_attributes_for :date_range
   end
 end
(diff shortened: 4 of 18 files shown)


spec/unit/app/models/block/edition_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    subject { concrete_edition_class.new(title: "Other Type Title") }
+    it { is_expected.to belong_to(:document).class_name("Block::Document") }
+  end
+
+  describe "validations" do
+    subject { concrete_edition_class.new(title: "Other Type Title") }
+    it { is_expected.to validate_presence_of(:title) }
+  end
+
+  describe "#to_details" do
```
edavey, line 68: nice, this is clearer than just changing the behaviour and our expectations of `#details`

app/models/block/time_period_edition.rb
```diff
@@ -0,0 +1,4 @@
+module Block
```
edavey, line 1: I think the commit message needs to be updated to refer to `#to_details` rather than `#details`?

app/models/block/edition.rb
```diff
@@ -0,0 +1,15 @@
+module Block
```
edavey, line 1: the commit msg refers to `#details` whereas we've actually implemented this as `#to_details` -- which I agree is an improvement...

app/models/block/time_period_date_range.rb
```diff
@@ -0,0 +1,32 @@
+module Block
```
edavey, line 1: I think we need to adjust the commit message as:

> Uses `datetime` columns for simple storage and validation (in contrast to
  the current main implementation which uses separate `date` and `time` fields.
  We expect the main current implementation to switch to using a single datetime field
  for `start` and for `end`)

refers to a situation which was true in the past but is now historic. I think we only need say something like:

> Uses `datetime` columns for simple storage and validation

spec/factories/block/time_period_date_range.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :block_time_period_date_range, class: "Block::TimePeriodDateRange" do
+    association :edition, factory: :block_time_period_edition
+    start { Time.zone.parse("2025-04-06 00:00") }
+    self.end { Time.zone.parse("2026-04-05 23:59") }
```
edavey, line 5: I think it's never that wise to use [keywords](https://docs.ruby-lang.org/en/2.2.0/keywords_rdoc.html) as field names, but perhaps we won't have too much aggro with `TimePeriodDateRange#end`?

spec/unit/app/models/block/time_period_date_range_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+        "start" => {
+          "date" => "2025-04-06",
+          "time" => "00:00",
+        },
+        "end" => {
+          "date" => "2026-04-05",
+          "time" => "23:59",
+        },
+      })
+    end
```
edavey, line 82: I think this is a description of how the date range _used_ to work? I think we now want `to_params` behaviour which is described something like:

```rb
{
  "date_range" => {
    "start" => "2028-04-06T00:00:00+01:00", 
    "end" => "2029-04-05T23:59:00+01:00"
  }
}
```

spec/unit/app/models/block/time_period_date_range_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    it "formats times with leading zeros" do
```
edavey: won't be needed I think?

spec/unit/app/models/block/time_period_edition_spec.rb
```diff
@@ -45,4 +45,59 @@
...
+        "date_range" => {
+          "start" => {
+            "date" => "2025-04-06",
+            "time" => "00:00",
+          },
+          "end" => {
+            "date" => "2026-04-05",
+            "time" => "23:59",
+          },
+        },
```
edavey, line 63: For `date_range` I think we're now expecting just `start` and `end` properties? Both in datetime form? Something like:

```
{
 "description" => "Tax year 2025/26",
  "date_range" => {
    "start" => "2025-04-06T00:00:00+01:00", 
    "end" => "2026-04-05T23:59:00+01:00"
  }
}
```

edavey (commented):
Looks good, thanks for this! 

Just: 

- a few tiny nitpicks e.g. where we refer to `details` rather than `to_details` in a few commit messages
- one issue where our new `TimePeriodDateRange#to_details` is representing separate date and time fields whereas I think in fact we've now simplified `TimePeriod#date_range` to use two datetimes (`start` and `end`)

spec/factories/block/time_period_date_range.rb
graycodes: Yeah I agree it's a bit of a pain using `self.` sporadically. I'm not super attached to it but the name itself is reasonable and I imagine that we're unlikely to do a lot of direct manipulation/access on `end` where it's not prefixed by something like `edition.date_range.end` etc.

spec/unit/app/models/block/time_period_date_range_spec.rb
graycodes: ah yep, removed now

app/models/block/document.rb
```diff
@@ -0,0 +1,30 @@
+module Block
+  class Document < ApplicationRecord
+    extend FriendlyId
+    friendly_id :sluggable_string, use: :slugged, slug_column: :content_id_alias, routes: :default
```
patrickjfl, line 4: do we need to defined `sluggable_string`?

app/models/block/document.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    has_many :editions, class_name: "Block::Edition", foreign_key: :block_document_id, dependent: :destroy
+
+    before_validation :generate_content_id, on: :create
+    after_validation :set_content_id_alias_and_embed_code, on: :create
+
+    def built_embed_code
```
patrickjfl, line 26: build rather than built?

app/models/block/document.rb
```diff
@@ -8,6 +8,10 @@ class Document < ApplicationRecord
     before_validation :generate_content_id, on: :create
     after_validation :set_content_id_alias_and_embed_code, on: :create
 
+    def title
+      editions.order(created_at: :desc).first&.title
+    end
```
patrickjfl, line 24: is this the existing behaviour in json schema world?

app/models/block/time_period_date_range.rb
```diff
@@ -0,0 +1,26 @@
...
+    belongs_to :edition, class_name: "Block::TimePeriodEdition"
+
+    validates :start, presence: true
+    validates :end, presence: true
+    validate :end_date_after_start_date
+
+    def to_details
+      {
+        "start" => start,
+        "end" => self.end,
```
patrickjfl, line 15: why `self.end`?

spec/unit/app/models/block/time_period_date_range_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        expect(date_range.errors[:end]).to include("must be after start date")
+      end
+
+      it "is valid when end is after start" do
+        date_range = described_class.new(
+          edition: edition,
+          start: Time.zone.parse("2025-04-06 00:00"),
+          end: Time.zone.parse("2026-04-05 23:59"),
+        )
+        expect(date_range).to be_valid
```
patrickjfl, line 66: think this one should be in its own describe block?

spec/unit/app/models/block/time_period_date_range_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    it { is_expected.to belong_to(:edition).class_name("Block::TimePeriodEdition") }
+  end
+
+  describe "validations" do
+    let(:edition) { create(:block_time_period_edition) }
+
+    subject { described_class.new(edition: edition, start: Time.zone.parse("2025-04-06 00:00"), end: Time.zone.parse("2026-04-05 23:59")) }
+
+    it { is_expected.to validate_presence_of(:start) }
+    it { is_expected.to validate_presence_of(:end) }
```
patrickjfl: why is this not in its own describe block? maybe it'd make sense alongside the 'valid date range' assertion

patrickjfl: also very minor but i think with the 23:59 / 00:00 setup i was expecting all valid or invalid times to centre around the hours and minutes (or date). so i had a moment of confusion before i saw the years had changed in the valid cases

app/models/block/document.rb
```diff
@@ -3,7 +3,18 @@ class Document < ApplicationRecord
...
+             foreign_key: :block_document_id,
+             dependent: :destroy,
+             inverse_of: :document
+
+    has_many :time_period_editions,
+             -> { where(type: "Block::TimePeriodEdition") },
+             class_name: "Block::TimePeriodEdition",
+             foreign_key: :block_document_id,
+             dependent: :destroy,
+             inverse_of: :document
```
patrickjfl, line 17: neat

spec/unit/app/models/block/time_period_date_range_spec.rb
```diff
@@ -11,6 +11,28 @@
     it { is_expected.to validate_presence_of(:start) }
     it { is_expected.to validate_presence_of(:end) }
 
+    describe "invalid date validation" do
+      it "is invalid when start date has invalid month" do
+        date_range = described_class.new(edition: edition)
+        # Simulate multiparameter assignment with invalid month (23)
+        date_range.start = { 1 => 2025, 2 => 23, 3 => 6, 4 => 0, 5 => 0 }
```
patrickjfl, line 20: is there a reason the invalid date is a complicated hash like this?

patrickjfl (dismissed):
Great work Graham, really chunky and sets us up well to continue the refactoring. I've just added a few minor comments and questions.

One issue is that all the commits seem to be coauthored by you and Ed, but for some reason Ed's part in that isn't verified, meaning that the commits aren't verified.

app/models/block/document.rb
graycodes: I could be wrong but we tend to favour these public methods to be named as nous rather than verbs, so the others are `title`, `embed_code_for_field` but I'm not an expert

app/models/block/document.rb
graycodes: Roughly. I've updated this on the branch you're now working on to use a `most_recent_edition.title` bit of sugar but it's about the same

app/models/block/time_period_date_range.rb
graycodes: annoyingly, end is a keyword in ruby so you can say `time_period_date_range.end` fine, but just calling `end` in this class appears like the end of a block so we say `self.end`

graycodes: Ed and I had a bit of a convo above ^ about changing it but I'm happy as-is for now

app/models/block/document.rb
graycodes: 🤔 hmm you might be right here. We possibly could have done without this for the first pass.

spec/unit/app/models/block/time_period_date_range_spec.rb
graycodes: I've changed up these dates a bit and put them in a describe to hopefully be a bit more clear

spec/unit/app/models/block/time_period_date_range_spec.rb
graycodes: I think as it's part of testing #end_date_after_start_date it should be here.

spec/unit/app/models/block/time_period_date_range_spec.rb
graycodes: 2 parts to this I guess.
1. Rails has this multipart date helper which allows users to send each part of a date as a separate field from the UI and it will automagically combine them back together. https://guides.rubyonrails.org/v3.2/form_helpers.html#select-model-object-helpers
2. We're specifically using that here as we're passing an invalid date. If we used `Time.zone.parse` with month set to 23 then it'd error.

app/models/block/time_period_date_range.rb
patrickjfl: ha, that is annoying!

spec/unit/app/models/block/time_period_date_range_spec.rb
patrickjfl: ah sorry, i was reading it as being within `invalid date validation`

spec/unit/app/models/block/time_period_date_range_spec.rb
patrickjfl: makes sense, thanks for explaining 👍

edavey (approved):

PR #778, opened 2026-07-13
Rename content modelling refs

edavey:
The "Content Modelling" team is being renamed to "Content Reuse".

This PR:

- renames the Slack channel `#govuk-publishing-content-{modelling,reuse}-dev`
- renames various references to "Content Modelling"

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,4 +125,4 @@ See the [`docs/`](docs/) directory.
 
 ## Contact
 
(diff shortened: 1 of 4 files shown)


graycodes (approved):

graycodes:
I've approved this here knowing that we haven't actually renamed the repo or the channel, but in expectation that we will shortly

PR #795, opened 2026-07-21
Remove postinstall from command form package.json

tobybatch:
This is causing a deploy error as tsc was missing from the deploy image. We decided to check in the pre-built files rather than adding tsc to the deploy image.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "license": "MIT",
   "packageManager": "yarn@1.22.22",
   "scripts": {
-    "postinstall": "cd node_modules/content-block-picker && npm ci && npm run build",
     "lint": "yarn run lint:js && yarn run lint:prettier && yarn run lint:scss && yarn run lint:markdown",
     "lint:fix": "yarn run lint:js:fix && yarn run lint:prettier:fix && yarn run lint:scss:fix && yarn run lint:markdown:fix",
     "lint:js": "eslint --cache --cache-location .cache/eslint --color -- \"**/*.js\"",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,6 +46,10 @@
     "js-yaml": "^4.2.0"
   },
   "dependencies": {
-    "content-block-picker": "alphagov/content-block-picker"
+    "content-block-picker": "alphagov/content-block-picker#main",
+    "govuk-frontend": "^6.4.0",
+    "js-yaml": "^4.2.0",
+    "jsdom": "^29.1.1",
+    "vite": "^8.1.5"
```
edavey: are you intending to make these other changes to `package.json` and `yarn.lock`?

The commit message only describes removing the `postinstall` cmd so I was half expecting to see just a single line removed? Probably it's my ignorance of the machinations of the JS toolchain but could you elaborate please as this perhaps a good opportunity to spread know-how by documenting background in our source control narrative?

yarn.lock
```diff
@@ -35,18 +35,18 @@
   integrity sha512-n8GuYSrI9bF7FFZ/SjhwevlHc8xaVlb/7HmHelnc/PZXBD2ZR49NnN9sMMuDdEGPeeRQ5d0hqlSlEpgCX3Wl0Q==
 
 "@babel/code-frame@^7.0.0":
-  version "7.29.0"
```
graycodes, line 38: I think we can completely remove the changes to this yarn.lock now as we're not updating any deps

tobybatch: I just ran yarn and let it sort the lock file. That babel entry shouldn't be anyting directly associated with the CPB demo.

graycodes: Ah yeah, I get the same locally so I guess these may be knock-ons from installing CBP 👍🏻

graycodes (approved):

PR #800, opened 2026-07-22
Use embed codes in initial contents of 'picker' demo

edavey:
By using some realistic copy which includes embed codes which are in use
in the integration environment, our users of the picker demo will be
immediately able to:

- see the highlighting of embed codes in action
- mouseover an embed code to view its rendered HTML

At present if the API responds with a 404 because we've tried to fetch
the rendered HTML for a embed code which can't be found, the picker throws
an error. Our cucumber specs currently specify that no JS errors are
encountered when the picker is loaded so perhaps this is suggesting
that an error is excessive for what's likely to be a routine occurrence?
-- when dealing with a free text textare perhaps it would be appropriate
to log a warning only on the client side?

For now we provide 2 sets of more realistic copy:

- with actual embed codes which work for the integration env
- with some realistic new state pension content in other envs

real_embed_code

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,5 +1,38 @@
 class PagesController < ApplicationController
   def accessibility_statement; end
 
-  def content_block_picker_demo; end
+  def content_block_picker_demo
+    block_picker_demo_initial_content
+  end
+
+private
+
+  def block_picker_demo_initial_content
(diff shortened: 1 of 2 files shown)


tobybatch (dismissed):
I did quite like the jargon jumble that was there, but I acknowledge yours is better.

tobybatch (approved):

PR #805, opened 2026-07-28
(CR-62) Create TimePeriodEdition

graycodes:
# Description

Allow users to create a new-style TimePeriodEdition.

Create standard form and fields to allow creation of a new time period edition, retaining much of the original functionality from the main app, including edition creation, GA4 attributes, JS modules, character count fields etc. Reuses existing Authoring functionality too.

Does not yet include the ability to Edit the edition or to navigate 'back' from this page, but this can follow shortly.

# Screenshot

Screenshot 2026-07-28 at 14 16 38

### DIFF
--- a/app/controllers/blocks_controller.rb
+++ b/app/controllers/blocks_controller.rb
@@ -1,0 +1,10 @@
+class BlocksController < ApplicationController
+  def edition_params
+    params.require("edition").permit(
+      :description,
+      :lead_organisation_id,
+      :instructions_to_publishers,
+      :title,
+    ).merge(creator: current_user)
+  end
+end
--- a/app/views/block/time_period_editions/new.html.erb
+++ b/app/views/block/time_period_editions/new.html.erb
@@ -1,6 +1,8 @@
 <% content_for :context, I18n.t("block/time_period_edition.new.context") %>
 <% content_for :title, I18n.t("block/time_period_edition.new.title") %>
 
+<% content_for :error_summary, render(Shared::ErrorSummaryComponent.new(object: @edition, parent_class: "edition")) %>
+
 <%= form_with(
   model: @edition,
   url: block_time_period_editions_path,
@@ -20,6 +22,7 @@
       hint: I18n.t("edition.hints.#{@edition.document.block_type}.title", default: nil),
       name: "edition[title]",
       value: @edition.title,
+      error_items: errors_for(@edition.errors, :title),
       rows: 1,
     },
     maxlength: 65,
@@ -41,6 +44,7 @@
   <%= render "govuk_publishing_components/components/select_with_search", {
     id: "edition_lead_organisation_id",
     name: "edition[lead_organisation_id]",
+    error_items: errors_for(@edition.errors, :lead_organisation_id),
     hint: I18n.t("edition.hints.not_displayed"),
     include_blank: true,
     label: I18n.t("edition.labels.lead_organisation"),
@@ -55,6 +59,7 @@
     hint: t("edition.hints.instructions_to_publishers_html"),
     textarea_id: "edition_instructions_to_publishers",
     value: @edition.instructions_to_publishers,
+    error_items: errors_for(@edition.errors, :instructions_to_publishers),
     } %>
 
   <div class="govuk-button-group govuk-!-margin-bottom-6">
--- a/app/views/block/time_period_editions/new.html.erb
+++ b/app/views/block/time_period_editions/new.html.erb
@@ -1,0 +1,67 @@
+<% content_for :context, I18n.t("block/time_period_edition.new.context") %>
+<% content_for :title, I18n.t("block/time_period_edition.new.title") %>
+
+<%= form_with(
+  model: @edition,
+  url: block_time_period_editions_path,
+  method: :post,
+  ) do |f| %>
+
+  <p class="govuk-body">
+    <%= I18n.t("edition.new.general_field_notice") %>
+  </p>
+
+  <%= render "govuk_publishing_components/components/character_count", {
+    id: "edition_title",
+    textarea: {
+      label: {
+        text: label_for_title(@edition.document.block_type),
+      },
+      hint: I18n.t("edition.hints.#{@edition.document.block_type}.title", default: nil),
+      name: "edition[title]",
+      value: @edition.title,
+      rows: 1,
+    },
+    maxlength: 65,
+    } %>
+
+  <div class="app-c-content-block-manager-textarea-component">
+    <%= render "components/govspeak_editor", {
+      name: "edition[description]",
+      id: "edition_description",
+      label: { text: I18n.t("edition.labels.description") },
+      value: @edition.description,
+      error_items: errors_for(@edition.errors, :description),
+      hint_id: "edition_description_hint",
+      rows: 5,
+      character_limit: 165,
+      } %>
+  </div>
+
+  <%= render "govuk_publishing_components/components/select_with_search", {
+    id: "edition_lead_organisation_id",
+    name: "edition[lead_organisation_id]",
+    hint: I18n.t("edition.hints.not_displayed"),
+    include_blank: true,
+    label: I18n.t("edition.labels.lead_organisation"),
+    options: taggable_organisations_container([@edition.lead_organisation_id]),
+    } %>
+
+  <%= render "govuk_publishing_components/components/textarea", {
+    label: {
+      text: I18n.t("edition.labels.instructions_to_publishers"),
+    },
+    name: "edition[instructions_to_publishers]",
+    hint: t("edition.hints.instructions_to_publishers_html"),
+    textarea_id: "edition_instructions_to_publishers",
+    value: @edition.instructions_to_publishers,
+    } %>
+
+  <div class="govuk-button-group govuk-!-margin-bottom-6">
+    <%= render "govuk_publishing_components/components/button", {
+      text: I18n.t("buttons.save_and_continue"),
+      type: "submit",
+      } %>
+  </div>
+
+<% end %>
--- a/app/views/block/time_period_editions/new.html.erb
+++ b/app/views/block/time_period_editions/new.html.erb
@@ -7,6 +7,7 @@
   model: @edition,
   url: block_time_period_editions_path,
   method: :post,
+  data: data_attributes_for_forms_with_text_fields(edition: @edition, block_type: @edition.document.block_type),
   ) do |f| %>
 
   <p class="govuk-body">
--- a/features/step_definitions/block/edition_steps.rb
+++ b/features/step_definitions/block/edition_steps.rb
@@ -1,0 +1,37 @@
+When("I am viewing the form to create a new Time Period Edition") do
+  visit new_block_time_period_edition_path
+end
+
+When("I fill the Time Period Edition details correctly") do
+  fill_in_edition_details(
+    title: "Test Time Period Edition",
+    description: "This is a test time period edition.",
+    lead_organisation: "Ministry of Example",
+    instructions: "Please review and publish.",
+  )
+end
+
+When("I fill the Time Period Edition details incorrectly") do
+  fill_in_edition_details(
+    title: "",
+    description: "This is a test time period edition.",
+    lead_organisation: "",
+    instructions: "Please review and publish.",
+  )
+end
+
+Then("I see that the Edition was created successfully") do
+  expect(page).to have_content(I18n.t("block/time_period_edition.create.success"))
+end
+
+Then("I see which Time Period Edition errors I need to correct") do
+  expect(page).to have_content("Title cannot be blank")
+  expect(page).to have_content("Lead organisation cannot be blank")
+end
+
+def fill_in_edition_details(title:, description:, lead_organisation:, instructions:)
+  fill_in "edition[title]", with: title
+  fill_in "Description", with: description
+  select lead_organisation, from: "Lead organisation"
+  fill_in "Instructions to publishers", with: instructions
+end
--- a/spec/controllers/blocks_controller_spec.rb
+++ b/spec/controllers/blocks_controller_spec.rb
@@ -1,0 +1,64 @@
+RSpec.describe BlocksController, type: :controller do
+  describe "#edition_params" do
+    let(:user) { create(:user) }
+
+    before do
+      allow(controller).to receive(:current_user).and_return(user)
+    end
+
+    context "when passed attributes" do
+      let(:raw_params) do
+        {
+          edition: {
+            lead_organisation_id: "123",
+            instructions_to_publishers: "Some instructions",
+            title: "A title",
+            unpermitted_field: "should be filtered out",
+          },
+        }
+      end
+
+      before do
+        controller.params = ActionController::Parameters.new(raw_params)
+      end
+
+      it "permits the expected top-level attributes" do
+        result = controller.send(:edition_params)
+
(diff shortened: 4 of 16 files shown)


spec/requests/block/time_period_editions_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.describe "Block time period editions", type: :request do
```
patrickjfl, line 1: nice use case 😎

app/views/block/time_period_editions/new.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    maxlength: 65,
+    } %>
+
+  <div class="app-c-content-block-manager-textarea-component">
+    <%= render "components/govspeak_editor", {
+      name: "edition[description]",
+      id: "edition_description",
+      label: { text: I18n.t("edition.labels.description") },
+      value: @edition.description,
+      hint_id: "edtion_description_hint",
```
patrickjfl: typo in `edition`

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -24,5 +24,27 @@
...
+        expect(form).to have_field("edition[description]")
+        expect(form).to have_field("edition[lead_organisation_id]")
+        expect(form).to have_field("edition[instructions_to_publishers]")
+        expect(form).to have_button("Save and continue")
+      end
+    end
+
+    it "initialises a time period edition with a time period document" do
+      get new_block_time_period_edition_path
+      edition = assigns(:edition)
```
patrickjfl: how does this work?

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -47,4 +47,52 @@
...
+      it "does not create a new time period edition and redirects to index" do
+        expect {
+          post block_time_period_editions_path, params: invalid_params
+        }.not_to change(Block::TimePeriodEdition, :count)
+
+        expect(response).not_to redirect_to(block_time_period_editions_path)
+      end
+
+      it "renders the new template with errors when invalid"
+    end
```
patrickjfl: is the implementation in the wrong commit?

spec/factories/block/time_period_edition.rb
```diff
@@ -4,5 +4,6 @@
     sequence(:title) { |n| "Time Period #{n}" }
     description { "A time period description" }
     lead_organisation_id { SecureRandom.uuid }
+    creator { association :user }
```
patrickjfl, line 7: this change might belong in its own commit

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -84,15 +84,24 @@
...
-      it "renders the new template with errors when invalid"
+      it "renders the new template with errors when invalid" do
+        post block_time_period_editions_path, params: invalid_params
+
+        expect(page).to have_css(".gem-c-error-summary__list-item", text: "Title cannot be blank")
+        expect(page).to have_css(".gem-c-error-summary__list-item", text: "Lead organisation cannot be blank")
+
+        expect(page).to have_css(".govuk-error-message", text: "Title cannot be blank")
+        expect(page).to have_css(".govuk-error-message", text: "Lead organisation cannot be blank")
+      end
```
patrickjfl, line 109: is there much value in having these two tests separate?

patrickjfl (dismissed):
Great work Graham, looks good to me. Just a few comments, mainly curiosity and git hygiene.

app/controllers/blocks_controller.rb
```diff
@@ -0,0 +1,9 @@
+class BlocksController < ApplicationController
+  def edition_params
+    params.require("edition").permit(
+      :lead_organisation_id,
+      :instructions_to_publishers,
+      :title,
```
edavey, line 7: I think we need to include `:description` here too? I noted that my description wasn't persisted and my log reports:

```
Unpermitted parameter: :description
```

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    context "when edition key is missing" do
+      before do
+        controller.params = ActionController::Parameters.new({})
+      end
+
+      it "raises ActionController::ParameterMissing" do
+        expect { controller.send(:edition_params) }
+          .to raise_error(ActionController::ParameterMissing)
```
edavey: should we go a bit further and assert that the error is indeed related to the lack of edition key? e.g.

```rb
expect { controller.send(:edition_params) }
  .to raise_error(
    ActionController::ParameterMissing,
    /param is missing or the value is empty or invalid: edition/,
)
```

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.describe "Block time period editions", type: :request do
+  before do
+    logout
+    login_as(create(:user))
+  end
+
+  describe "GET /block/time_period_editions" do
```
edavey: do we have a use-case for showing all time period editions in a list?

If it's 

> 3. A page to redirect the user to, once successful.

would we be better off using a "member" page e.g. `GET /block/time_period_editions/{:id}`? That might be a tidier pathway:

- use it initially for the redirection after creation of `TimePeriodEdition`, before `TimePeriodDateRange` is implemented
- have a long term use for displaying a particular edition of a document

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -24,5 +24,27 @@
...
+    end
+
+    it "initialises a time period edition with a time period document" do
+      get new_block_time_period_edition_path
+      edition = assigns(:edition)
+
+      expect(edition).to be_a(Block::TimePeriodEdition)
+      expect(edition).to be_new_record
+      expect(edition.document).to be_new_record
+      expect(edition.document.block_type).to eq("time_period")
```
edavey: nice!

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -47,4 +47,52 @@
...
+          },
+        }
+      end
+
+      it "creates a new time period edition and redirects to index" do
+        expect {
+          post block_time_period_editions_path, params: valid_params
+        }.to change(Block::TimePeriodEdition, :count).by(1)
+
+        expect(response).to redirect_to(block_time_period_editions_path)
```
edavey: perhaps we could also assert that the expected flash msg is in place?

e.g. this could become something like

> "creates a new time period edition and redirects to index with success message"

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -47,4 +47,52 @@
...
+
+      it "does not create a new time period edition and redirects to index" do
+        expect {
+          post block_time_period_editions_path, params: invalid_params
+        }.not_to change(Block::TimePeriodEdition, :count)
+
+        expect(response).not_to redirect_to(block_time_period_editions_path)
+      end
+
+      it "renders the new template with errors when invalid"
```
edavey: did you mean this 

> "renders the new template with errors when invalid"

stub to be committed here? Perhaps it belongs where in a later commit where `create` gains some conditional behaviour based on validation?

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -84,15 +84,24 @@
...
+      it "renders the new template with errors when invalid" do
+        post block_time_period_editions_path, params: invalid_params
+
+        expect(page).to have_css(".gem-c-error-summary__list-item", text: "Title cannot be blank")
+        expect(page).to have_css(".gem-c-error-summary__list-item", text: "Lead organisation cannot be blank")
+
+        expect(page).to have_css(".govuk-error-message", text: "Title cannot be blank")
+        expect(page).to have_css(".govuk-error-message", text: "Lead organisation cannot be blank")
+      end
     end
```
edavey, line 110: hmm, I think the naming  and organisation of these two examples is a little muddled?
```
context "with invalid parameters"

  it "does not create a new time period edition and re-renders the form with errors"
    -> asserts 
        - nothing created
        - http status
        - template name
        
  it "renders the new template with errors when invalid" do
    -> asserts 
        - error messages

```

Perhaps they should be reorganised or decomposed, e.g.:

```
context "when invalid parameters are supplied"

  it "does not create a time period edition"
    asserts:
      - nothing persisted
      - http status

  it "re-renders the form with error messages"
    asserts:
      - template name
      - error messages
```

features/step_definitions/block/edition_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+Then("I see which Time Period Edition errors I need to correct") do
+  expect(page).to have_content("Title cannot be blank")
+  expect(page).to have_content("Lead organisation cannot be blank")
+end
+
+def fill_in_edition_details(title:, description:, lead_organisation:, instructions:)
+  fill_in "edition[title]", with: title
+  fill_in "Description", with: description
+  select lead_organisation, from: "Lead organisation"
+  fill_in "Instructions to publishers", with: instructions
```
edavey, line 36: I really appreciate how simple these steps are 👍

app/views/block/time_period_editions/new.html.erb
```diff
@@ -7,6 +7,7 @@
   model: @edition,
   url: block_time_period_editions_path,
   method: :post,
+  data: data_attributes_for_forms_with_text_fields(edition: @edition, block_type: @edition.document.block_type),
```
edavey, line 10: where should we describe this behaviour in our test suite? In the controller test? We're already making asserting about the contents on the page when describing our error messages when validation fails on `:create`. But that is probably not a great place for describing something so generic?

More long-term we could use:
- a "view" as we do in `spec/views/design_system_spec.rb` (e.g. a "blocks form"  view spec which asserts on every form in our `block` space somehow?)
- have a "request" spec which describes general characteristics of the HTML/DOM/content. We do that for the accessibility statement (spec/requests/pages_spec.rb)

But if it's appropriate to add the `data_attributes_for_forms_with_text_fields` helper here then I think it also must be appropriate to describe its presence and prevent regression?

app/views/block/time_period_editions/new.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    textarea: {
+      label: {
+        text: label_for_title(@edition.document.block_type),
+      },
+      hint: I18n.t("edition.hints.#{@edition.document.block_type}.title", default: nil),
+      name: "edition[title]",
+      value: @edition.title,
+      rows: 1,
+    },
+    maxlength: 65,
```
edavey, line 29: how are we going to describe this character-count behaviour? Would it be sufficient at this point (rather than demonstrating the behaviour) to assert that the DOM includes the expected attributes/content? e.g.

-  css of `textarea.govuk-js-character-count`
- hidden `#edition_title_info` of "You can enter up to 65 characters"

edavey (commented):
great work! Just a few minor comments/suggestions.

spec/requests/block/time_period_editions_spec.rb
graycodes: This comes from a gem called `rails-controller-testing` which implements `assigns(:foo)` as a way of getting access to the instance var `@foo`.

See the readme here https://github.com/rails/rails-controller-testing#usage

spec/requests/block/time_period_editions_spec.rb
graycodes: I left it here just to show my intent to add it later but you've both commented that it looks weird so I've taken it back out!

spec/requests/block/time_period_editions_spec.rb
graycodes: Fair enough. I just wanted a throwaway page to show a confirmation but I've changed it to be something useful (unlike my useless one). It now shows the `/block/time_period_editions/{:id}` page instead

spec/requests/block/time_period_editions_spec.rb
graycodes: I left it here just to show my intent to add it later but you've both commented that it looks weird so I've taken it back out!

spec/requests/block/time_period_editions_spec.rb
graycodes: Oh yeah this is a good point. I will collapse these together!

app/views/block/time_period_editions/new.html.erb
graycodes: Ah yeah I agree this seems like a good idea. I've added a `:view` test to check for the presence of these with a decent level of specificity which I think covers us here.

app/views/block/time_period_editions/new.html.erb
graycodes: I've now created another test to check for the presence of these fields so that we know if there are any regressions.

spec/views/block/time_period_edition_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      expect(rendered).to have_css('form[data-module~="unsaved-changes-prompt"]')
+    end
+
+    it "should attach the required data attributes to enable the 'ga4 form tracking' for the user" do
+      render
+
+      expect(rendered).to have_css('form[data-ga4-action="create"]')
+      expect(rendered).to have_css('form[data-module~="ga4-form-tracker"]')
+      expect(rendered).to have_css('form[data-ga4-tool-name="time_period"]')
+    end
```
edavey, line 21: works well!

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+
+        expect(response).to have_http_status(:unprocessable_content)
+        expect(response).to render_template("block/time_period_editions/new")
+
+        expect(page).to have_css(".gem-c-error-summary__list-item", text: "Title cannot be blank")
+        expect(page).to have_css(".gem-c-error-summary__list-item", text: "Lead organisation cannot be blank")
+
+        expect(page).to have_css(".govuk-error-message", text: "Title cannot be blank")
+        expect(page).to have_css(".govuk-error-message", text: "Lead organisation cannot be blank")
+      end
```
edavey, line 109: Nit-picky: all in one spec works too, but might be worth wrapping in an `aggregate_failures` block?

spec/requests/block/time_period_editions_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        end
+      end
+
+      it "contains character-limited fields with the correct character limits" do
+        expect(page).to have_css(".govuk-character-count[data-module='govuk-character-count'][data-maxlength='65']")
+        expect(page).to have_css(".govuk-character-count__message", text: "You can enter up to 65 characters")
+
+        expect(page).to have_css(".govuk-character-count[data-module='govuk-character-count'][data-maxlength='165']")
+        expect(page).to have_css(".govuk-character-count__message", text: "You can enter up to 165 characters")
+      end
```
edavey, line 45: neat.

edavey (approved):
Great stuff, thanks for all the little tweaks.

spec/requests/block/time_period_editions_spec.rb
patrickjfl: neat. thanks!

spec/requests/block/time_period_editions_spec.rb
graycodes: ah yes good idea! Added

graycodes: ah yes good idea! Added

PR #809, opened 2026-07-29
Add an index page for viewing ActiveRecord documents

patrickjfl:
This PR adds a new index page for viewing refactored ActiveRecord documents. This is a copy of the existing index page for the JSON schema documents, albeit at this point it does not include filter functionality - that will come in a subsequent PR.

Rather than adding endless conditional statements to the existing summary card component so that it’s also compatible with the ActiveRecord document, I’ve created a ‘v2’ summary card component, which only serves ActiveRecord documents. This should make it much easier and cleaner to turn off the v1 implementation and move over entirely to v2, when the time comes.

Screenshot 2026-07-29 at 15 36 04

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,13 +113,13 @@ and running.
 To run all linters, run:
 
 ```bash
-bundle exec rake:lint
+bundle exec rake lint
 ```
 
 To run all linter with automatic formatting corrections, run:
 
 ```bash
-bundle exec rake:lint_autocorrect
+bundle exec rake lint_autocorrect
 ```
 
 ### API
--- a/app/components/document/index/v2/summary_card_component.rb
+++ b/app/components/document/index/v2/summary_card_component.rb
@@ -1,0 +1,55 @@
+class Document::Index::V2::SummaryCardComponent < ViewComponent::Base
+  def initialize(document:)
+    @document = document
+  end
+
+private
+
+  attr_reader :document
+
+  def rows
+    [
+      title_item,
+      description_item,
+      organisation_item,
+    ].compact
+  end
+
+  def title_item
+    {
+      key: helpers.label_for_title(document.block_type),
+      value: document.title,
+    }
+  end
+
+  def description_item
+    {
+      key: "Description",
+      value: edition.description,
+    }
+  end
+
+  def organisation_item
+    {
+      key: "Lead organisation",
+      value: edition.lead_organisation.name,
+    }
+  end
+
+  def title
+    document.title
+  end
+
+  def summary_card_actions
+    [
+      {
+        label: "View",
+        href: "",
+      },
+    ]
+  end
+
+  def edition
+    @edition = document.most_recent_edition
+  end
+end
--- a/app/models/block/document.rb
+++ b/app/models/block/document.rb
@@ -26,6 +26,16 @@ module Block
 
     enum :block_type, { time_period: "time_period" }
 
+    scope :by_most_recently_created_edition, lambda {
+      latest_edition = Block::Edition
+        .select(:created_at)
+        .where("block_editions.block_document_id = block_documents.id")
+        .order(created_at: :desc)
+        .limit(1)
+
+      order(Arel.sql("(#{latest_edition.to_sql}) DESC NULLS LAST"))
+    }
+
     def title
       most_recent_edition&.title
     end
--- a/app/models/block/document.rb
+++ b/app/models/block/document.rb
@@ -16,13 +16,18 @@ module Block
              dependent: :destroy,
              inverse_of: :document
 
+    has_one :most_recent_edition,
+            -> { most_recent_first },
+            foreign_key: :block_document_id,
+            class_name: "Block::Edition"
+
     before_validation :generate_content_id, on: :create
     after_validation :set_content_id_alias_and_embed_code, on: :create
 
     enum :block_type, { time_period: "time_period" }
 
     def title
-      editions.order(created_at: :desc).first&.title
+      most_recent_edition&.title
     end
 
     def built_embed_code
--- a/features/block/view_content_blocks.feature
+++ b/features/block/view_content_blocks.feature
@@ -1,0 +1,13 @@
+Feature: Search for a content object
+  Background:
+    Given I am logged in
+    And the organisation "Department of Placeholder" exists
+    And the organisation "Ministry of Example" exists
+    And the following time period content blocks have been drafted:
+        | Time period name              | Description                             | Organisation              |
+        | British Summer Time           | The period of daylight savings          | Department of Placeholder |
+        | Student loan repayment period | The repayment period for a student loan | Ministry of Example       |
+
+Scenario: view all available content blocks
+    When I visit the blocks index page
+    Then I should see the details for all three available content blocks
--- a/spec/components/document/index/v2/summary_card_component_spec.rb
+++ b/spec/components/document/index/v2/summary_card_component_spec.rb
@@ -1,0 +1,40 @@
+RSpec.describe Document::Index::V2::SummaryCardComponent, type: :component do
+  include Rails.application.routes.url_helpers
+
+  let(:document) { build_stubbed(:block_document, block_type: "time_period") }
+  let(:organisation) { build(:organisation) }
+
+  let(:edition) do
+    build_stubbed(
+      :block_time_period_edition,
+      id: 123,
+      description: "some time period description",
+      lead_organisation_id: organisation.id,
+      updated_at: 1.day.ago,
+      document: document,
+    )
+  end
+
+  before do
+    expect(document).to receive(:most_recent_edition).at_least(:once).and_return(edition)
+    allow(Organisation).to receive(:all).and_return([organisation])
+  end
+
+  it "renders a content block as a summary card" do
+    render_inline(described_class.new(document:))
+
+    expect(page).to have_css ".govuk-summary-card__title", text: edition.title
+    expect(page).to have_css ".govuk-summary-card__action", count: 1
+    expect(page).to have_css ".govuk-summary-card__action .govuk-link[href='']"
+
+    expect(page).to have_css ".govuk-link", text: "View"
+
+    expect(page).to have_css ".govuk-summary-list__row", count: 3
+
+    expect(page).to have_summary_row.with_key("Time period name").with_value(edition.title)
+
+    expect(page).to have_summary_row.with_key("Description").with_value(edition.description)
+
+    expect(page).to have_summary_row.with_key("Lead organisation").with_value(edition.lead_organisation.name)
+  end
+end
--- a/spec/requests/block/documents_spec.rb
+++ b/spec/requests/block/documents_spec.rb
@@ -1,0 +1,42 @@
+RSpec.describe "Block documents", type: :request do
+  before do
+    logout
+    login_as(create(:user))
+  end
+
+  describe "GET /block/documents" do
+    let(:organisation) { build(:organisation) }
+
+    before do
+      allow(Organisation).to receive(:all).and_return([organisation])
+    end
+
+    it "renders the index page with listed block documents sorted by most recently created edition" do
+      document_with_older_edition = create(:block_document)
+      create(
+        :block_time_period_edition,
+        document: document_with_older_edition,
+        title: "Tax Year 2025",
+        lead_organisation_id: organisation.id,
+        created_at: 1.year.ago,
+      )
+
+      document_with_newer_edition = create(:block_document)
+      create(
+        :block_time_period_edition,
+        document: document_with_newer_edition,
+        title: "British Summer Time 2026",
+        lead_organisation_id: organisation.id,
+        created_at: 1.day.ago,
+      )
+
+      get block_documents_path
+
+      expect(response).to have_http_status(:success)
+      expect(response).to render_template("block/documents/index")
+
+      expect(page).to have_css("[data-testid='homepage-item-0']", text: "British Summer Time 2026")
+      expect(page).to have_css("[data-testid='homepage-item-1']", text: "Tax Year 2025")
+    end
+  end
+end
--- a/spec/unit/app/models/block/document_spec.rb
+++ b/spec/unit/app/models/block/document_spec.rb
@@ -95,6 +95,20 @@ RSpec.describe Block::Document, type: :model do
     end
   end
 
+  describe ".by_most_recently_created_edition" do
+    it "orders documents so the document with the most recently created edition appears first" do
+      older_document_with_newer_edition = create(:block_document, created_at: 4.days.ago)
+      create(:block_time_period_edition, document: older_document_with_newer_edition, created_at: 4.days.ago)
+      create(:block_time_period_edition, document: older_document_with_newer_edition, created_at: 1.day.ago)
+
+      newer_document_with_older_edition = create(:block_document, created_at: 3.days.ago)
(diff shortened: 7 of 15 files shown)


spec/components/document/index/summary_card_component_v2_spec.rb
```diff
@@ -0,0 +1,47 @@
+RSpec.describe Document::Index::SummaryCardComponentV2, type: :component do
+  include Rails.application.routes.url_helpers
+
+  include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::SanitizeHelper
+  include EditionHelper
```
graycodes: I'm not sure you need any of these includes. Was there something they were needed for?

app/components/document/index/v2/summary_card_component.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def title
+    document.title
+  end
+
+  def summary_card_actions
+    [
+      {
+        label: "View",
+        href: "",
```
graycodes, line 47: Is it worth marking this as some kind of todo here or in the commit message?

app/models/block/document.rb
```diff
@@ -24,6 +24,8 @@ class Document < ApplicationRecord
     before_validation :generate_content_id, on: :create
     after_validation :set_content_id_alias_and_embed_code, on: :create
 
+    scope :by_most_recently_updated, -> { order(updated_at: :desc) }
```
graycodes: 🤔 based on our recent conversation, should this be the document with the most recently updated edition?

spec/requests/block/documents_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      get block_documents_path
+
+      expect(response).to have_http_status(:success)
+      expect(response).to render_template("block/documents/index")
+
+      html = Nokogiri::HTML(response.body)
+
+      expect(html.at_css("[data-testid='homepage-item-0']").text).to include("Tax Year 2026")
+      expect(html.at_css("[data-testid='homepage-item-1']").text).to include("Tax Year 2025")
```
graycodes: I think rspec will give you `page` which is a Capybara node so you can do stuff like 

```rb
      expect(page).to have_css("[data-testid='homepage-item-0']", text: "Tax Year 2026")
      expect(page).to have_css("[data-testid='homepage-item-1']", text: "Tax Year 2025")
```
without the need to invoke nokogiri

graycodes (commented):
I am tempted to say that the Block namespace might be preferable for the component as you mentioned

spec/components/document/index/summary_card_component_v2_spec.rb
patrickjfl: good spot! thanks, unnecessary bloat copied over from the v1 file

app/components/document/index/v2/summary_card_component.rb
patrickjfl: i woke up with this very thought this morning, yes absolutely

app/models/block/document.rb
patrickjfl: hmmmm, i sort of thought that's what i was doing, but maybe that reveals my unit tests aren't up to scratch. looks like I can achieve that by leaving this as it is and adding `touch: true` to the Edition model? https://guides.rubyonrails.org/association_basics.html#touch

spec/requests/block/documents_spec.rb
patrickjfl: ah thank you! 🛤️

app/models/block/document.rb
patrickjfl: In line with what we decided for which document edition should take precedence (most recently created), I've gone with the same approach here, ordering by most recently created edition. It's a bit more cumbersome than keeping the original scope and adding `touch: true` to the Edition model, but more consistent.

app/models/block/document.rb
```diff
@@ -24,6 +24,16 @@ class Document < ApplicationRecord
...
     after_validation :set_content_id_alias_and_embed_code, on: :create
 
+    scope :by_most_recently_created_edition, lambda {
+      latest_edition = Block::Edition
+        .select(:created_at)
+        .where("block_editions.block_document_id = block_documents.id")
+        .order(created_at: :desc)
+        .limit(1)
+
+      order(Arel.sql("(#{latest_edition.to_sql}) DESC NULLS LAST"))
```
graycodes, line 36: I think we may be able to simplify this subquery by using a MAX and ordering on that. Specifying the `where` you have here is possibly overkill for rails as you can just do `left_outer_joins(:editions)` to magically use its rels.

I think this gives the same result:
```sql
  left_outer_joins(:editions)
    .select('block_documents.*, MAX(block_editions.created_at) AS recent_edition_created_at')
    .group('block_documents.id')
    .order('recent_edition_created_at DESC NULLS LAST')
```

graycodes (dismissed):
I've left a comment about sql queries but I think regardless this should have an approval. Looks good.

app/models/block/document.rb
patrickjfl: I've put these two options to Gemini, which reckons the subquery is the better approach, primarily because 1) it preserves `.count` and pagination, and 2) as our list of editions grow, it will be more performant, provided we add the following composite index to the Edition database: `add_index :block_editions, [:block_document_id, :created_at]`, which will allow the scope to pick out the most recently created edition for each document rather than sorting through all of them. What do you reckon?

graycodes: Ah okay fair enough. I'm not sure what you mean about preserving count and pagination though. If you get a minute to explain that I'd appreciate it. 
I was mostly thinking about readability here. At our scales I think these queries are going to be trivial anyway but if my suggestion causes issues then by all means go with what you have.

spec/unit/app/models/block/document_spec.rb
```diff
@@ -32,6 +32,22 @@ def lead_organisation_id
         expect(document.time_period_editions).not_to include(other)
       end
     end
+
+    describe "#most_recent_edition" do
+      it "returns the most recently created edition" do
+        document = create(:block_document, block_type: "time_period")
+        _older_edition = create(:block_time_period_edition, document: document, created_at: 2.days.ago)
+        newer_edition = create(:block_time_period_edition, document: document, created_at: 1.day.ago)
```
edavey, line 43: When describing ordering it's often more reliable to locate the fixture you're asserting on in the middle of the sequence of creation, e.g.

- `expected_not_first`
- `expected_first`
- `expected_not_first`

That way you know that the ordering isn't working by chance due to the creation order or default ordering... (i.e. consider having 3 fixtures)

spec/components/document/index/v2/summary_card_component_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(page).to have_css ".govuk-summary-card__title", text: edition.title
+    expect(page).to have_css ".govuk-summary-card__action", count: 1
+    expect(page).to have_css ".govuk-summary-card__action .govuk-link[href='']"
+
+    expect(page).to have_css ".govuk-link", text: "View"
+
+    expect(page).to have_css ".govuk-summary-list__row", count: 3
+
+    expect(page).to have_css ".govuk-summary-list__key", text: "Time period name"
+    expect(page).to have_css ".govuk-summary-list__value", text: edition.title
```
edavey: We have a neat [`have_summary_row` custom matcher ](https://github.com/alphagov/govuk-content-block-manager/blob/8699e6a1f7c3f2339d62b2d521679d1ea944fd1a/spec/support/matchers/have_summary_row.rb#L1)which we could use for a more expressive description like:

```rb
expect(page).to have_summary_row.with_key("Title").with_value(edition.title)
```

spec/unit/app/models/block/document_spec.rb
```diff
@@ -91,6 +91,18 @@ def lead_organisation_id
...
+    it "orders documents so the document with the most recently created edition appears first" do
+      document_with_older_edition = create(:block_document)
+      create(:block_time_period_edition, document: document_with_older_edition, created_at: 2.days.ago)
+
+      document_with_newer_edition = create(:block_document)
+      create(:block_time_period_edition, document: document_with_newer_edition, created_at: 1.hour.ago)
+
+      expect(described_class.by_most_recently_created_edition).to eq([document_with_newer_edition, document_with_older_edition])
+    end
+  end
```
edavey, line 110: In order to describe the grouping behaviour thoroughly here do you need more than 1 edition fixture per document e.g. given:

- older_document
- newer_document

where:

- older doc has editions from Mon and Thurs
- newer doc has editions from Tues and Weds

-> we expect older doc is first in list

i.e. to describe the sub-query?

```sql
SELECT "block_documents".* 
FROM "block_documents" 
ORDER BY (
	SELECT "block_editions"."created_at" FROM "block_editions" 
	WHERE (block_editions.block_document_id = block_documents.id) 
	ORDER BY "block_editions"."created_at" DESC LIMIT 1
) DESC NULLS LAST
```

spec/requests/block/documents_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      )
+
+      document_with_newer_edition = create(:block_document)
+      create(
+        :block_time_period_edition,
+        document: document_with_newer_edition,
+        title: "Tax Year 2026",
+        lead_organisation_id: organisation.id,
+        created_at: 1.day.ago,
+      )
```
edavey, line 31: I think this is a little confusing as an example. When the tax year changes we will create a new edition of the same document. We won't create a new document.

I think it would be good to have two editions of the same tax year document:

- 2025-26
- 2026-27

and assert that we see the 2026-27 title

and then have another document time period doc (nothing to do with tax year) with an edition which is either earlier or later than the 2026-27 edition and then assert its position relative to 2026-27.

i.e. I think there are two behaviours to describe:

1. how documents are ranked according to the date of their most recent edition
2. how each document shown presents the title of its most recent edition

features/block/view_content_blocks.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Search for a content object
+  Background:
+    Given I am logged in
+    And the organisation "Department of Placeholder" exists
+    And the organisation "Ministry of Example" exists
+    And the following time period content blocks have been drafted:
+        | Time period name              | Description                             | Organisation              |
+        | 2026-27 tax year              | The tax year starting on 1st April 2026 | Department of Placeholder |
+        | 2025-26 tax year              | The tax year starting on 1st April 2027 | Department of Placeholder |
```
edavey: As we saw in the unit tests, creating a new document for a subsequent edition is wrong. This could be rather misleading and confusing!

Probably it would be more realistic to do something like 

- create 2 documents (e.g. the tax year and British Summer Time )
- create 3 editions 

assert that we see 2 editions. Assuming that the unit tests describe the ordering and selection of the particular editions, you could limit your assertion to something like

> Then I see the details for both available content blocks

README.md
```diff
@@ -113,13 +113,13 @@ and running.
...
-bundle exec rake:lint
+bundle exec rake lint
 ```
 
 To run all linter with automatic formatting corrections, run:
 
 ```bash
-bundle exec rake:lint_autocorrect
+bundle exec rake lint_autocorrect
 ```
```
edavey, line 123: thanks for spotting and fixing up

edavey (commented):
A few minor suggestions and comments. 

And a more major documentation issue with creating new tax year documents when the correct pattern is to make a "further edition" of the document which is in use. (A new document means a new embed code, which means no ability to update the documents using the original block...)

app/models/block/document.rb
patrickjfl: sorry, yes. if you run `Block::Document.by_most_recently_created_edition.count` on my version, you get a number, but if you run it on yours, you get an error: `ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near "AS" (ActiveRecord::StatementInvalid)
LINE 1: ...block_documents.*, MAX(block_editions.created_at) AS recent_...`.
Apparently that's because when you run `.count` on a query, Rails takes what's inside your `.select()` statement and wraps it in a count function, which causes a PostgreSQL error because it doesn't like 'AS' aliases inside a count function.

graycodes: Oh wow! I had no idea. That's a good thing to know though going forward.

spec/unit/app/models/block/document_spec.rb
patrickjfl: could you show me what you mean here for this test? I'm not quite following. Thanks

spec/components/document/index/v2/summary_card_component_spec.rb
patrickjfl: thanks!

spec/unit/app/models/block/document_spec.rb
patrickjfl: yes good point, have amended 👍

spec/requests/block/documents_spec.rb
patrickjfl: yes I think that's a fair challenge, I will update the test so the example causes less confusion. Personally though I'm not sure we need to test `2. how each document shown presents the title of its most recent edition` in this file, as we're already testing the `title` method in the model tests: https://github.com/alphagov/content-block-manager/blob/426756202c5453a51ed44d22bc6b0f84c2023a31/spec/unit/app/models/block/document_spec.rb#L129. To me that seems the obvious place to test what the title of a document refers to. What do you think?

features/block/view_content_blocks.feature
patrickjfl: ok yes, I'll slim it down down to two examples and make sure they aren't both tax years.

spec/unit/app/models/block/document_spec.rb
edavey:
```rb
it "returns the most recently created edition" do
  document = create(:block_document, block_type: "time_period")

  _older_edition = create(:block_time_period_edition, document: document, created_at: 2.days.ago)
  newer_edition = create(:block_time_period_edition, document: document, created_at: 1.day.ago)
  _oldest_edition = create(:block_time_period_edition, document: document, created_at: 3.days.ago)

  expect(document.most_recent_edition).to eq(newer_edition)
end
```

We create our fixtures in a sequence. They might come out of the db in a particular order based on the creation sequence. By positioning our target fixture (`newer_edition`, the one on whose position we're going to assert) in the middle of the creation sequence we know that it's not ending up first or last in the query's ordering as a result of how the fixtures were set up. Just a small point. I hope that's clearer now?

patrickjfl: aha, yes, with you now, that make total sense. thanks for explaining, i'll update that.

spec/requests/block/documents_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      get block_documents_path
+
+      expect(response).to have_http_status(:success)
+      expect(response).to render_template("block/documents/index")
+
+      html = Capybara.string(response.body)
```
graycodes: You can just use `page` here e.g.
```rb
      expect(page).to have_css("[data-testid='homepage-item-0']", text: "British Summer Time 2026")
```

graycodes (dismissed):

spec/requests/block/documents_spec.rb
patrickjfl: oh yes, thanks

graycodes (approved):

PR #821, opened 2026-08-03
Stop publishing blocks to Content Store

edavey:
The process of conforming to [RFC-192][rfc_192] brought it to our attention that we don't in fact currently have a valid use-case for publishing our blocks to the Content Store and therefore don't need to publish a base_path. See [ADR 15: Remove dependence on base_path][adr_15].

In this commit we remove the two elements of the "publish" payload which cause a document to be lodged at the Content Store:

- `base_path`
- `routes`

See how "downstream jobs" update the Content Store only if `base_path` is present:

- [downstream draft job][draft]
- [downstream live job][live]

[rfc_192]:
https://github.com/alphagov/govuk-rfcs/pull/192/

[adr_15]:
https://github.com/alphagov/content-block-manager/blob/ea52adea1dcaeaede9b3530295abe08ae045c676/docs/architecture/decisions/0015-remove-dependence-on-base-path.md

[draft]:
https://github.com/alphagov/publishing-api/blob/42eb06a8a67d59c2f142fa9cc72d424c57c8b60c/app/sidekiq/downstream_draft_job.rb#L40

[live]:
https://github.com/alphagov/publishing-api/blob/42eb06a8a67d59c2f142fa9cc72d424c57c8b60c/app/sidekiq/downstream_live_job.rb#L43

### DIFF
--- a/app/presenters/publishing_api/content_block_presenter.rb
+++ b/app/presenters/publishing_api/content_block_presenter.rb
@@ -14,12 +14,10 @@ module PublishingApi
         title: edition.title,
         instructions_to_publishers: edition.instructions_to_publishers,
         content_id_alias:,
-        base_path:,
         details: edition.details,
         links:,
         update_type:,
         change_note:,
-        routes:,
       }
     end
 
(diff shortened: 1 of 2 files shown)


graycodes (approved):
Looks good!

PR #829, opened 2026-08-04
(CR-90) Rename Block -> V2

graycodes:
# Description

Rename all instances referencing `Block` and `block_...` to
reference the new V2 namespace.

The name `Block` was problematic to use in this project. The name is
generic and doesn't convey much meaning, even when compared to the
existing non-Block code instances, for example the models:

- `Document` vs `Block::Document`
- `Edition` vs `Block::Edition`

These names don't adequately describe the difference between the need
for two similarly named models.

In addition, the word 'block' is already overloaded in this repo which
provides users the ability to manage content blocks.

As such, we choose the new namespace `V2` in order to more accurately
describe our intent, as can be seen when comparing the new models to
the existing ones, e.g.:

- `Document` vs `V2::Document`
- `Edition` vs `V2::Edition`

This signifies our intent to build a new version of existing work,
achieving similar goals but with a different approach.

## Jira Ticket

https://gov-uk.atlassian.net/browse/CR-90


edavey (dismissed):
Very neat! 🚀

app/models/v2/document.rb
```diff
@@ -1,35 +1,35 @@
...
              foreign_key: :block_document_id,
              dependent: :destroy,
              inverse_of: :document
 
     has_one :most_recent_edition,
             -> { most_recent_first },
-            foreign_key: :block_document_id,
-            class_name: "Block::Edition"
+            foreign_key: :v2_document_id,
+            class_name: "V2::Edition"
```
patrickjfl, line 22: should this change be in the next commit?

app/models/v2/document.rb
```diff
@@ -1,35 +1,35 @@
...
     after_validation :set_content_id_alias_and_embed_code, on: :create
 
     enum :block_type, { time_period: "time_period" }
 
     scope :by_most_recently_created_edition, lambda {
-      latest_edition = Block::Edition
+      latest_edition = V2::Edition
         .select(:created_at)
-        .where("block_editions.block_document_id = block_documents.id")
+        .where("v2_editions.v2_document_id = v2_documents.id")
```
patrickjfl, line 32: same here

db/migrate/20260803145359_rename_block_tables_to_v2.rb
```diff
@@ -0,0 +1,8 @@
+class RenameBlockTablesToV2 < ActiveRecord::Migration[8.1]
+  def change
+    rename_table :block_documents, :v2_documents
+    rename_table :block_editions, :v2_editions
+    rename_column :v2_editions, :block_document_id, :v2_document_id
+    rename_table :block_time_period_date_ranges, :v2_time_period_date_ranges
+  end
+end
```
patrickjfl, line 8: does it make sense to have this work in a separate commit from the first? Is either commit atomic without the other?

patrickjfl (commented):
The changes look good to me, thanks very much for doing all this awkward work Graham. I've got one question about the commit history. Also I'm wondering if it would make sense to update [ADR 16](https://github.com/alphagov/content-block-manager/blob/main/docs/architecture/decisions/0016-store-blocks-as-activerecord-models.md), replacing the references to the `Block` namespace with `V2`. I think it has no effect on the meaning of the ADR but may clear up future confusion.

app/models/v2/document.rb
graycodes: Ooh yeah good spot, you're right. Will move now

db/migrate/20260803145359_rename_block_tables_to_v2.rb
graycodes: They do work atomically, as this is just a behind the scenes change. So, the tests pass and the code works, but the class names are mismatched with their table names. I feel like as the tests this is probably alright

patrickjfl: the tests pass, but if you have existing blocks in your DB and check out and run the code in the previous commit, you would see errors without the migration in this commit, right?

graycodes: If I understand correctly, no. We'd be using the class `V2::Document` which would point at the table `block_documents`. We'd join on `block_editions` using the `block_document_id`. 

At the point where we do the table/column rename, we also tell ActiveRecord to use the newly named tables and columns. 

That said, if I'm wrong about this then shout

patrickjfl: Ok fair enough, I think you're right. It feels a bit odd in Rails where so much is based on conventional file and class name matching to have the class names come apart from the tables, but if we're explicitly pointing to the existing table names and columns like you say then it shouldn't break anything. I would probably put it all in the same commit, but happy to go with this approach too 👍

patrickjfl (approved):
Great work Graham

PR #831, opened 2026-08-05
Stub SchedulePublishingWorker dequeue method in workflow_spec

patrickjfl:
This test has been failing locally for me as Sidekiq has attempted and failed to connect to a Redis server in the test run. That connection has nothing to do with what’s being asserted in the test, so we can safely stub out SchedulePublishingWorker’s dequeue method.

### DIFF
--- a/spec/requests/workflow_spec.rb
+++ b/spec/requests/workflow_spec.rb
@@ -709,6 +709,7 @@ RSpec.describe "Workflow", type: :request do
       describe "#update" do
         describe "when choosing to publish immediately" do
           it "redirects to the review step" do
+            allow(SchedulePublishingWorker).to receive(:dequeue)
             scheduled_at = {
               "scheduled_publication(1i)": "",
               "scheduled_publication(2i)": "",


graycodes (approved):
The tests still seem to pass here and also locally so I think this looks good, thanks! 

I do see a similar message in the `Editions::StatusTransitionsController` but I assume this isn't failing for you locally?

```
Editions::StatusTransitionsController
  #create
    when the valid transition is 'publish'
      retrieves the given edition
      attempts to transition the given edition to the relevant state
      when the transition succeeds
        redirects to the correct page
        shows a success message
    when the valid transition is 'delete'
2011-11-11T11:11:11.000Z pid=73260 tid=1njo INFO: Sidekiq 7.3.10 connecting to Redis with options {size: 10, pool_name: "internal", url: "redis://127.0.0.1:6379", reconnect_attempts: [0.05, 0.25, 1, 5, 15, 30]}
      retrieves the given edition
      attempts to transition the given edition to the relevant state
```

spec/requests/workflow_spec.rb
```diff
@@ -709,6 +709,7 @@ def self.it_shows_the_correct_context
       describe "#update" do
         describe "when choosing to publish immediately" do
           it "redirects to the review step" do
+            allow(SchedulePublishingWorker).to receive(:dequeue)
```
edavey, line 712: if you were to put the stub into a before block, it might be more obvious that this has nothing to do with our description of `#update` e.g.

```rb
describe "#update" do
  before { allow(SchedulePublishingWorker).to receive(:dequeue) }
  
  describe "when choosing to publish immediately" do
    it "redirects to the review step" do
      scheduled_at = {
        "scheduled_publication(1i)": "",
        "scheduled_publication(2i)": "",
        "scheduled_publication(3i)": "",
        "scheduled_publication(4i)": "",
        "scheduled_publication(5i)": "",
      }

      put workflow_path(id: edition.id, step:),
          params: {
            schedule_publishing: "now",
            scheduled_at:,
          }

      expect(response).to redirect_to(workflow_path(id: edition.id, step: :review))
    end
  end
```

The dequeue behaviour **is** described in `spec/unit/app/services/publish_edition_service_spec.rb` I see...

PR #835, opened 2026-08-07
(CR-91) Edit Time Period Edition

graycodes:
# Description

Previous work introduced a page to create a new Time Period Edition. 
This builds on that work, reusing the form to allow the user to edit their
existing TPEs.

As of yet, there is no navigation to these pages but we can make this 
available in future work.

# Jira ticket
https://gov-uk.atlassian.net/browse/CR-91


edavey (approved):
Nice work, very clear!

Just a suggestion for next time: perhaps it might be helpful to use the testability of elements as a guide to the granularity of commits? I find it's quite an advantage to aim for each commit to include a "description" of the changes made in the form of changes/additions to the automated tests. In that way the documentation and proof of correctness advance in lock-step with the implementation. That can often be a useful driver towards the "right" granularity of commit. 

For example: here, I think you might present the work in fewer commits, each with their own test element:

- successfully update a `TimePeriodEdition` with valid params (with Rspec examples)
- fail to update a `TimePeriodEdition` with invalid params (with Rspec examples)
- less detailed descriptions of both happy and sad paths in cucumber features

If you wish to define the routing separately for clarity (I do find that helpful to break out) you could do that with an initial commit:
- add 'edit' and 'update' routes (with [routing specs](https://rspec.info/features/6-0/rspec-rails/routing-specs/))

Just food for thought. 

But I'm always uneasy when I see a commit which doesn't include any test: I immediately ask myself whether it could/should have a test or whether it ought to be part of larger commit which is tested...

PR #855, opened 2026-08-18
CR-98: allow users to create new date range

patrickjfl:
Allows users to add a date range to a refactored, v2 time period edition.

Credit to @edavey for a good portion of the implementation, from his original [experimental branch.
](https://github.com/alphagov/content-block-manager/pull/553)
https://github.com/user-attachments/assets/74dc13d8-b9bb-41a5-be02-2157747e87f5

### DIFF
--- a/app/controllers/v2/time_period_date_ranges_controller.rb
+++ b/app/controllers/v2/time_period_date_ranges_controller.rb
@@ -1,0 +1,16 @@
+module V2
+  class TimePeriodDateRangesController < ApplicationController
+    before_action :set_edition
+
+    def edit; end
+
+    def update; end
+
+  private
+
+    def set_edition
+      @edition = @document.time_period_editions.find(params[:id])
+    end
+
+  end
+end
--- a/app/controllers/v2/time_period_date_ranges_controller.rb
+++ b/app/controllers/v2/time_period_date_ranges_controller.rb
@@ -4,13 +4,31 @@ module V2
 
     def edit; end
 
-    def update; end
+    def update
+      if @edition.update(edition_params)
+        redirect_to v2_time_period_edition_path(@edition.id),
+                    notice: I18n.t("v2/time_period_edition.create.success")
+      else
+        @error_summary_errors = @edition.errors.map do |error|
+          {
+            text: error.message,
+            href: "##{error.attribute.to_s.tr('.', '_')}",
+          }
+        end
+        render :edit, status: :unprocessable_content
+      end
+    end
 
   private
 
     def set_edition
-      @edition = @document.time_period_editions.find(params[:id])
+      @edition = V2::TimePeriodEdition.find(params[:time_period_edition_id])
     end
 
+    def edition_params
+      params.require(:edition).permit(
+        date_range_attributes: %i[id start end],
+      )
+    end
   end
 end
--- a/app/views/v2/time_period_editions/edit.html.erb
+++ b/app/views/v2/time_period_editions/edit.html.erb
@@ -1,4 +1,4 @@
-<% content_for :context, I18n.t("block/time_period_edition.edit.context") %>
-<% content_for :title, I18n.t("block/time_period_edition.edit.title") %>
+<% content_for :context, I18n.t("v2/time_period_edition.edit.context") %>
+<% content_for :title, I18n.t("v2/time_period_edition.edit.title") %>
 
 <%= render "form", url: v2_time_period_edition_path, method: :put %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -49,6 +49,7 @@ en:
             lead_organisation_id:
               blank: "cannot be blank"
         v2/time_period_date_range:
+          format: "%{message}"
           attributes:
             start:
               blank: "Enter a start date"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -435,7 +435,7 @@ en:
       hour: Hour
       minute: Minute
 
-  block/time_period_edition:
+  v2/time_period_edition:
     new:
       title: "Create time period"
       context: "Create content block"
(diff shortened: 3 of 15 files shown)


config/routes.rb
```diff
@@ -47,7 +47,11 @@
 
   namespace :v2 do
     resources :time_period_editions, only: %i[new create show edit update]
-    resources :documents, only: [:index]
+    resources :documents, only: [:index] do
+      resources :time_period_date_ranges,
```
edavey: would this be clearer as a [singular route](https://guides.rubyonrails.org/routing.html#singular-resources), as we can only every have a maximum of one date range per time period?

```rb
resource :time_period_date_range,
			   path: "time-period-date-range"
 
```

config/locales/en.yml
```diff
@@ -49,6 +49,7 @@ en:
             lead_organisation_id:
               blank: "cannot be blank"
         v2/time_period_date_range:
+          format: "%{message}"
```
edavey: interesting! I don't remember being aware of this before, though I see [we're using it on the edition model](https://github.com/alphagov/content-block-manager/blob/2f30075539d8b8b2065b826f19b067bbc585eba5/config/locales/en.yml#L15).

Is it worth including a link to docs where this feature is described? e.g. https://tekin.co.uk/2026/07/ten-cool-things-you-might-not-know-about-rails-i18n

spec/requests/v2/time_period_date_ranges_spec.rb
```diff
@@ -0,0 +1,81 @@
+RSpec.describe "V2::TimePeriodDateRanges", type: :request do
+  setup do
+    logout
+    user = create(:user)
+    login_as(user)
+  end
+
+  describe "PATCH /v2/documents/:document_id/time-period-date-ranges/:id" do
```
edavey: as suggested earlier, would 

```
PATCH /v2/documents/:document_id/time-period-date-range/:id
```
be clearer?

edavey (dismissed):
👍 looking good, just a couple of minor suggestions.

What's next up? "Editor views time period including date range after editing"?

config/routes.rb
patrickjfl: yes I think you're right 👍

config/locales/en.yml
patrickjfl: thanks, have included

spec/requests/v2/time_period_date_ranges_spec.rb
patrickjfl: yes, agreed and implemented

graycodes (dismissed):

config/routes.rb
```diff
@@ -47,7 +47,11 @@
 
   namespace :v2 do
     resources :time_period_editions, only: %i[new create show edit update]
-    resources :documents, only: [:index]
+    resources :documents, only: [:index] do
+      resources :time_period_date_range,
+                only: %i[edit update],
+                path: "time-period-date-range"
```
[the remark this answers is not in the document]
patrickjfl: well, because we want it to be nested under the parent model in the path, which also means the parent ID will be passed in as a parameter and available in the controller. Although you raise a good point, I've amended it so that it's nested under the time period edition, which is its true parent (and also changed `resources` to `resource` seeing as we are expecting only one date range per edition).

app/controllers/v2/time_period_date_ranges_controller.rb
```diff
@@ -5,7 +5,20 @@ class TimePeriodDateRangeController < ApplicationController
...
-    def update; end
+    def update
+      if @edition.update(edition_params)
+        redirect_to v2_time_period_edition_path(@edition.id),
+                    notice: I18n.t("v2/time_period_edition.create.success")
+      else
+        @error_summary_errors = @edition.errors.map do |error|
+          {
+            text: error.message,
+            href: "##{error.attribute.to_s.tr('.', '_')}",
```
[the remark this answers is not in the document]
patrickjfl: I was having all kinds of problems linking to the fields from the error summary, because the attribute was `date_range.start` rather than `date_range_start`. This ensures the href matches the field name in the view.

app/controllers/v2/time_period_date_range_controller.rb
```diff
@@ -5,7 +5,20 @@ class TimePeriodDateRangeController < ApplicationController
...
+        redirect_to v2_time_period_edition_path(@edition.id),
+                    notice: I18n.t("v2/time_period_edition.create.success")
+      else
+        @error_summary_errors = @edition.errors.map do |error|
+          {
+            text: error.message,
+            href: "##{error.attribute.to_s.tr('.', '_')}",
+          }
+        end
+        render :edit, status: :unprocessable_entity
```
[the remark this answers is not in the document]
patrickjfl: thanks!

app/views/v2/time_period_editions/edit.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :context, I18n.t("block/time_period_edition.edit.context") %>
-<% content_for :title, I18n.t("block/time_period_edition.edit.title") %>
+<% content_for :context, I18n.t("v2/time_period_edition.edit.context") %>
+<% content_for :title, I18n.t("v2/time_period_edition.edit.title") %>
```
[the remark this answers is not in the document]
patrickjfl: ah yes good spot!

app/controllers/v2/time_period_date_ranges_controller.rb
```diff
@@ -5,7 +5,20 @@ class TimePeriodDateRangeController < ApplicationController
 
     def edit; end
 
-    def update; end
+    def update
```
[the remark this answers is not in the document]
patrickjfl: agreed! credit to @edavey

graycodes (approved):

PR #858, opened 2026-08-19
(CR-45) Pull content block picker from NPM

graycodes:
Pull content-block-picker from NPM.

Slight tweaks to the way it's imported to add it to the `application.js` bundle with the rest of our UMD modules.

Removes the inline script tag as it's no longer needed.


patrickjfl (approved):
LGTM 👍

PR #862, opened 2026-08-20
Include directions to republish blocks when changed

edavey:
We had a minor incident following our change to the structure of pension blocks to [present currency symbol independent of the stored amount][pr_689]. 

1. we failed to republish changed blocks to the Publishing API

2. host documents which include these blocks, when *they* were re-presented  to the Content Store used the stale content block `#details`. This resulted in seeing amounts formatted with two £ symbols e.g."££193.47".

The fix was to republish the 2 blocks affected and then observe how through [dependency resolution at Publishing API][dependency_resolution] all the host documents using these two content blocks were re-presented to the Content Store using the correct content block values.

In this commit we update the guidance for making changes to blocks to emphasise the need to publish changed blocks to the Publishing API.

[pr_689]:
https://github.com/alphagov/content-block-manager/pull/689

[dependency_resolution]:
https://github.com/alphagov/publishing-api/blob/01a0c466f173f535f1f098af61fe22e3579c4421/docs/dependency-resolution.md

### DIFF
--- a/docs/changing_the_structure_of_a_content_block.md
+++ b/docs/changing_the_structure_of_a_content_block.md
@@ -2,6 +2,22 @@
 
 If you need to change the structure of a content block, you will need to coordinate your changes across the CBM repo, the GOVUK E2E test repo, the Content Reuse E2E test repo, as well as running any relevant data migrations in each environment. You may also need to introduce defensive changes to Content Block Tools and CBM so ensure that both the legacy and the new data structures can be accommodated adequately whilst the changes are in flight.
 
+## Republishing changed blocks
+
+Content blocks whose `#details` have changed must be republished to the Publishing API so that host documents are re-presented to the Content Store using the new form of the block. This re-presentation happens automatically at the Publishing API through its [dependency resolution system][dependency_resolution]. Documents identified as being "dependencies" of the changed block are sent "downstream" by the [`DependencyResolutionJob`][dependency_resolution_job].
+
+If changed blocks are not republished by Content Block Manager then when a host document is re-presented to the Content Store either:
+
+- by a Content Manager from a publishing app such as Whitehall or Mainstream Publisher, or
+- as part of a bulk re-presentation from Publishing API for maintenance reasons
+
+then the outdated block lodged at the Publishing API will be used in the host document when [translating embed codes to HTML][content_embed_presenter].  
+
+Options for republishing from Content Block Manager:
+
+- where there are small numbers of blocks concerned, this can be done via the UI (the "Edit block" journey makes a new edition and republishes it)
+- once there are large quantities of blocks to republish, a rake task can be used to create new editions and publish them with `PublishEditionService.new.call(edition)`
+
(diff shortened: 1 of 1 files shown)


graycodes (approved):

docs/changing_the_structure_of_a_content_block.md
```diff
@@ -2,6 +2,22 @@
 
 If you need to change the structure of a content block, you will need to coordinate your changes across the CBM repo, the GOVUK E2E test repo, the Content Reuse E2E test repo, as well as running any relevant data migrations in each environment. You may also need to introduce defensive changes to Content Block Tools and CBM so ensure that both the legacy and the new data structures can be accommodated adequately whilst the changes are in flight.
 
+## Republishing changed blocks
+
+Content blocks whose `#details` have changed must be republished to the Publishing API so that host documents are re-presented to the Content Store using the new form of the block. This re-presentation happens automatically at the Publishing API through its [dependency resolution system][dependency_resolution]. Documents identified as being "dependencies" of the changed block are sent "downstream" by the [`DependencyResolutionJob`][dependency_resolution_job].
```
patrickjfl: does this only apply when `#details` are changed?

docs/changing_the_structure_of_a_content_block.md
```diff
@@ -2,6 +2,22 @@
...
+
+If changed blocks are not republished by Content Block Manager then when a host document is re-presented to the Content Store either:
+
+- by a Content Manager from a publishing app such as Whitehall or Mainstream Publisher, or
+- as part of a bulk re-presentation from Publishing API for maintenance reasons
+
+then the outdated block lodged at the Publishing API will be used in the host document when [translating embed codes to HTML][content_embed_presenter].
```
patrickjfl: I'm finding this construction a bit hard to follow. How about something like:

```
A host document may be re-presented to the Content Store either:
- by a Content Manager from a publishing app such as Whitehall or Mainstream Publisher, or
- as part of a bulk re-presentation from Publishing API for maintenance reasons.
If when this happens changed blocks have not been republished by Content Block Manager, then the outdated block lodged at the Publishing API will be used in the host document when [translating embed codes to HTML][content_embed_presenter].
```

docs/changing_the_structure_of_a_content_block.md
edavey: I've changed this to read 

>Content blocks which have been changed must be republished

This documentation `changing_the_structure_of_a_content_block.md` is concerned with changing the structure of the block (the `#details`) but for example if the `#change_note` or `#instructions_to_publishers` had been changed then republishing to Publishing API is also necessary.

docs/changing_the_structure_of_a_content_block.md
edavey: thanks for recasting this in a clearer form. Much better!

patrickjfl (approved):
Looks good, thanks for adding this documentation.

PR #863, opened 2026-08-20
(CR-145) Ignore slack links in markdown when linting

graycodes:
Slack has introduced browser detection in their links, meaning that when the markdown-link-check tool runs it receives a 403 status code when checking links to slack channels. This marks the link as broken.

As such, the easy fix is to ignore links to slack when doing these checks. This does mean that if this link really becomes dead then we may not know.

I tried allowing cookies to be saved locally when accessing these links programmatically but this didn’t seem to fix it. I’m not sure how exactly they are calculating if a user is using a browser but I haven't found a way around it.

# Jira Ticket

https://gov-uk.atlassian.net/browse/CR-145

### DIFF
--- a/markdown-link-check-config.json
+++ b/markdown-link-check-config.json
@@ -5,6 +5,9 @@
     },
     {
       "pattern": "^https?://localhost"
+    },
+    {
+      "pattern": "https?://.*.slack.com/archives/.*"
     }
   ]
 }


edavey:
I feel we ought to include the more detailed information from the Jira ticket here in the **commit message.** 

i.e. I'm proposing that our git history contain all the necessary narrative for a full understanding of code changes and should not rely on links to other resources like Jira and Slack. These can and do become unavailable for all sorts of reasons. (As this commit itself attests!). From Jira CR-145:

> We have been using markdown-link-checker to ensure that none of the links in our markdown files are broken. This can happen when someone enters an invalid link or a link goes dead due to changes in the host application.
> 
> Today, Slack implemented browser detection in their links to slack channels. Now if you hit one of their links programmatically it returns a 403 (such as with curl or node). When following a link in a browser it still works. Links returning a 403 are considered ‘broken’.
> 
> I tried allowing cookies to be saved locally when accessing these links programmatically but this didn’t seem to fix it. I’m not sure how exactly they are calculating if a user is using a browser but I can’t find a way around it.
> 
> # Outcome
> 
> As such I think we should just ignore checking markdown links that link to slack.

edavey (approved):

